PR #1, opened 2020-10-14
Ensure last effect step generates enough samples.

microbit-matt-hillsdon:


finneyj:
Thanks @microbit-matt-hillsdon, good spot. There's nothing like a good rounding error to start off a new repo. :)

PR #2, opened 2020-10-19
SoundExpressions::play is now sync.

microbit-matt-hillsdon:
Adds event in SoundEmojiSynthesizer so we can detect when play is complete.


finneyj:
Thanks @microbit-matt-hillsdon. Looks good!

PR #3, opened 2020-10-24
Saving flash

martinwork:
SoftDevice S113
bootloader.o compiled for S113 and address 0x77000, with uicr.o.
Updated LD
FDS and MicroBitStorage locations swapped.

This PR needs to be merged at the same time as https://github.com/microbit-foundation/codal-microbit-nrf5sdk/pull/11

The bootloader comes from https://github.com/microbit-foundation/v2-bootloader/pull/14

See https://github.com/microbit-foundation/codal-microbit/issues/137


finneyj:
thanks @martinwork!

Issue #6, opened 2020-10-28
Ability to stop a sound expression


pelikhan:
We need to be able to button-smash a button and play the latest sound expression.

jaustin:
This is nice for ‘button bashing’ and avoids queues. It’s also the behaviour I’d expect if we were playing back a sample so you can make a sampler that plays back the ‘sample’ ’s s s s sample sam sam sample’ if you press the button at the right time.

finneyj:
But you might also want the queuing behaviour... The ability to back-to-back sounds also has applications.

Would it make sense to follow the design pattern we have for other long running tasks, like scrolling/animating the display:
https://github.com/lancaster-university/codal-core/blob/master/source/drivers/AnimatedDisplay.cpp#L252

Here we have an explicit abort... the cancels any active animation and moves onto the next one. 

I'm not sure I agree this behaviour is what you'd expect for playing back a sample. it' snot like you say 

print ("Jonny");
print ("Jonny");
print ("Jonny");

and expect J-J-Jonny... 

:)

pelikhan:
Having a block to abort any sound works too.

________________________________

finneyj:
Support for this added as of v0.2.15...
https://github.com/lancaster-university/codal-microbit-v2/blob/master/source/SoundExpressions.cpp#L103

PR #7, opened 2020-10-28
Remove locking from SoundEmojiSynthesizer.

microbit-matt-hillsdon:
Playing a new sound changes what's being played.
The sync play method on sound expressions still allows waiting after
playing.

One approach to #6.

### DIFF
--- a/inc/SoundEmojiSynthesizer.h
+++ b/inc/SoundEmojiSynthesizer.h
@@ -109,7 +109,6 @@ namespace codal
         public:
 
         DataSink*               downStream;             // Our downstream component.
-        FiberLock               lock;                   // Ingress queue to handle concurrent playback requests.
         ManagedBuffer           buffer;                 // Current playout buffer.
         ManagedBuffer           effectBuffer;           // Current sound effect sequence being generated.
         ManagedBuffer           emptyBuffer;            // Zero length buffer.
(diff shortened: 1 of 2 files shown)


microbit-matt-hillsdon:
This simple version it would mean that sync play could also be interrupted. And a constant stream of sounds would mean that an interrupted sound expression sync play might have its event repeatedly deferred. 

If that isn't desirable I think we’d need to push the sync/async code into the SoundEmojiSynthesizer. It’s simple approach though.

finneyj:
Hmm... Fully agree that having a way to stop playback of a sound effect is desirable but I'm not totally convinced of the approach. Simply because SoundExpressions isn't the only consumer of this API...

microbit-matt-hillsdon:
@finneyj, who wants the locking? Would it be easy for them to wait for the event in the same way as sound expressions sync play, which gets you the same behaviour as the locking (I think) ?

Completely happy to go with another approach though, I just thought I'd try this as it seemed straightforward.

finneyj:
hmmm.. currently only users are SoundExpressions and virtualOutputPin I think. We'll likely have a MakeCode editor onto that API in the future but not yet. I thought micropython also had a direct API on it, but actually that goes straight to the mixer. 

I think I'm just a bit cautious here as this is not really what the rest of the API looks like. All other calls are blocking unless you call an explicit Async flavour. and we already have a similar patter for the display (when you scroll text etc), so it feels to me like we should be consistent...

microbit-matt-hillsdon:
Closing in favour of #9.

PR #9, opened 2020-10-30
Add support for stopping sound expressions.

microbit-matt-hillsdon:
Alternative to #7 to address #6.

See https://github.com/lancaster-university/microbit-v2-samples/pull/5 for a test that lets you explore the behaviour.

### DIFF
--- a/inc/SoundEmojiSynthesizer.h
+++ b/inc/SoundEmojiSynthesizer.h
@@ -128,7 +128,7 @@ namespace codal
         int                     samplesWritten;         // The number of samples written from the current sound effect block.
         float                   position;               // Position within the tonePrint.
         float                   samplesPerStep[EMOJI_SYNTHESIZER_TONE_EFFECTS];     // The number of samples to render per step for each effect.
-
+        bool                    stopping;               // Tracks when a stop is requested until it is acted on in a pull.
         /**
           * Default Constructor.
           * Creates an empty DataStream.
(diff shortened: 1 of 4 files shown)


inc/SoundEmojiSynthesizer.h
```diff
@@ -128,7 +128,7 @@ namespace codal
         int                     samplesWritten;         // The number of samples written from the current sound effect block.
         float                   position;               // Position within the tonePrint.
         float                   samplesPerStep[EMOJI_SYNTHESIZER_TONE_EFFECTS];     // The number of samples to render per step for each effect.
-
+        bool                    stopping;               // Tracks when a stop is requested until it is acted on in a pull.
```
microbit-matt-hillsdon, line 131: I could use a status flag instead?

Also not sure whether all these fields should be public, but I've followed that pattern.

finneyj: I agree, a lot of these should be private really. Looks like an oversight. Let's fix that in a different commit. bool is fine, but optimising to a status flag would be nice to save a few bytes though :)

finneyj:
Thanks @microbit-matt-hillsdon. Looks grand!

inc/SoundEmojiSynthesizer.h
microbit-matt-hillsdon: Status flag done in https://github.com/lancaster-university/codal-microbit-v2/pull/10

PR #10, opened 2020-10-30
Optimisation: Use a flag for stopping state.

microbit-matt-hillsdon:


finneyj:
nice - thanks @microbit-matt-hillsdon!

PR #13, opened 2020-11-03
Polyfilling mbed for extensions

microbit-sam:
Implements: PwmOut, InterruptIn, DigitalIn, DigitalOut, Ticker, (Mbed)Timer

Extensions wishing to use any of these will need to make sure that they `#include "mbed.h"`

Is there a define extensions can use determine which timer to use? e.g. `Timer` in on v1, and `MbedTimer` on v2
```
#ifdef IS_V2
MbedTimer timer;
#elseif
Timer timer;
#endif
```

Note: commit messages referencing extensions were chosen as they were top of this list rather than any favoritism!
https://github.com/microsoft/pxt-microbit/issues/3506

@finneyj please critique!


inc/compat/MbedTimer.h
```diff
@@ -0,0 +1,60 @@
+#ifndef MbedTimer_h
+#define MbedTimer_h
+
+class MbedTimer {
+    private:
+        uint32_t _start;
+        uint32_t _stop = -1;
```
microbit-sam, line 7: This is causing my timer trouble!

microbit-sam:
This doesn't compile properly in PXT..

```
error: reference to 'uBit' is ambiguous
   39 |             uBit.messageBus.listen(DEVICE_ID_MBED_TICKER, 0x0, this, &Ticker::onTick);
      |             ^~~~

~/pxt-microbit-next/projects/Untitled/built/codal/pxtapp/pxt.h:46:17: note: candidates are: 'codal::MicroBit pxt::uBit'
   46 | extern MicroBit uBit;
```

microbit-sam:
Fixed by using `__has_include` to include `pxt.h` if it exists

I can replace it with a check for a define if `__has_include` isn't supported in all environments? Couldn't find a `PXT_EXISTS` type define anywhere

microbit-sam:
Now uses `MICROBIT_DAL_FIBER_USER_DATA` to determine if it is in PXT or not

finneyj:
Thanks Sam - this looks really good. nice job - thanks!

PR #23, opened 2020-11-10
#define Timer MbedTimer

microbit-sam:
`mbed.h` will only be included by files that don't know about CODAL, right?


finneyj:
yep.

microbit-sam:
Ok cool, please can we add this

pelikhan (approved):

finneyj:
thanks @microbit-sam

PR #24, opened 2020-11-12
Off chip storage

finneyj:
Would like to merge this in ASAP so that we can:

 - unify APIs for on chip and off chip flash storage
 - save 4K of flash (using the upstream codal KeyValueStorgae class doesn't need a scratch page)
 - enable on chip storage applications for v2 (compass calibration etc).

@microbit-sam Can you please see minor changes to PartialFlashing and let me know if this is OK. Not additional parameter now needed to constructor.

@martinwork Can you please review the similar changes to BLE init and let me know if these are of any concern.


martinwork:
I guess bleManager doesn't need to be in the initialiser list to ensure its constructor gets called before interrupts are enabled?

It looks like you've forgotten to change where BLEManager creates PartialFlashingService.

finneyj:
Thanks @martinwork - great catch. I forgot to set that CONFIG option in my test.

I'm not quite sure what you're referring to about the bleManager / initialization list though. Where are you referring to exactly?

martinwork:
I just noticed that bleManager is no longer in the MicroBit::MicroBit initialiser list:
https://github.com/lancaster-university/codal-microbit-v2/blob/f5c2c9edcdc981ffc447f154679ed8e5b144e1e6/model/MicroBit.cpp#L53

though it's still at the top here:
https://github.com/lancaster-university/codal-microbit-v2/blob/f5c2c9edcdc981ffc447f154679ed8e5b144e1e6/model/MicroBit.h#L126

I'm not sure why bleManager is removed, when some simple constructors are still included, like:
https://github.com/lancaster-university/codal-microbit-v2/blob/f5c2c9edcdc981ffc447f154679ed8e5b144e1e6/model/MicroBit.cpp#L81

finneyj:
Ahh, I see what you mean now. Thanks @martinwork. 

I think simply because we were now using the default constructor, so an explicit call was unnecessary. Agree it is good to include for consistency though.

finneyj:
@microbit-sam - Any thoughts on those minor changes to PartialFlashing before I merge?

microbit-sam:
No, looks good to me!

finneyj:
thanks @microbit-sam @martinwork

PR #43, opened 2020-12-02
Fix BLE disconnection

martinwork:


finneyj:
Thanks @martinwork 

Can you please just update the comment indicating which IRQ priorities are available when SoftDevice is enabled, and we can merge in?

martinwork:
Corrected.

finneyj:
ta!

PR #51, opened 2020-12-11
Allow more than one base UUID to be used in a service

martinwork:
The MakeCode HF2 service uses different vendor specific base UUIDs for its service and characteristic.


martinwork:
Reduce ram origin #52 includes this change and can be pulled instead.

finneyj:
ok - thanks @martinwork. Let's just close this off then in favour of #52. I'm sure we'll merge shortly, as soon as you've convinced me that we're not double allocating memory :)

PR #64, opened 2021-01-04
BLE Partial Flashing for MicroPython

microbit-sam:


source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -329,17 +329,16 @@ void MicroBitPartialFlashingService::partialFlashingEvent(MicroBitEvent e)
       // Iterate through until reaching the scratch page
       while(flashPointer < (uint32_t *)DEFAULT_SCRATCH_PAGE)
       {
-        // Check for embedded source magic
-        if(*flashPointer == 0x41140EF && *(uint32_t *)(flashPointer + 0x1) == 0xB82FA2BB)
+
+        // Check if page is blank
+        if(crc32_compute((uint8_t*)flashPointer, MICROBIT_CODEPAGESIZE, NULL) != 0xc71c0011)
```
microbit-sam, line 334: Erases remaining pages in region for both MicroPython and MakeCode, so it will remove the embedded source if it is there

source/MicroBitMemoryMap.cpp
```diff
@@ -138,5 +140,58 @@ void MicroBitMemoryMap::findHashes()
...
+            // Found uPy Magic
+            DMESG("FOUND UPY_MAGIC");
+            uint8_t nRegions = (*(magicAddress - (sizeof(magicAddress)/2)));
+            
+            // Iterate through regions
+            for(int i = 0; i < nRegions; i++) {
+               processRecord(magicAddress - ((2 + i) * sizeof(magicAddress))); 
+            }
+
+        }
```
microbit-carlos, line 156: With all this pointer arithmetic maybe it's getting to the point where it might be more clear if there was a struct defined with the different entries being members of the expected size.
We can probably do the same with the record rows.

source/MicroBitMemoryMap.cpp
```diff
@@ -138,5 +140,58 @@ void MicroBitMemoryMap::findHashes()
...
+    start = start * 4096;
+
+    uint32_t length; memcpy(&length, address + 1, 4); 
+
+    uint32_t hash[2]; 
+    if(hashType == 2) {
+        memcpy(&hash, (uint32_t *) *(address + 2), 8);
+    } else {
+        memcpy(&hash, address + 2, 8);
+    }
```
microbit-carlos, line 179: It'd be good to add a note about what the different hash types are, unless it's already covered somewhere else.
An enum with the note would be ideal.

source/bluetooth/MicroBitPartialFlashingService.cpp
microbit-carlos: I guess this means MakeCode partial flashing will now be slightly slower?

microbit-carlos: Also, should the changes to this file be backported to the DAL as well?

source/MicroBitMemoryMap.cpp
```diff
@@ -138,5 +140,58 @@ void MicroBitMemoryMap::findHashes()
...
+/*
+ * Function to process record from uPy build
+ */
+void MicroBitMemoryMap::processRecord(uint32_t *address) {
+    uint8_t id = (*address) - 1;
+    uint8_t hashType = ((*address) >> 8) & 0xFF;
+
+    uint32_t start; memcpy(&start, address, 4); 
+    start = (start & 0xFFFF0000) >> 8;
+    start = start * 4096;
```
microbit-carlos, line 170: Can we use MICROBIT_CODEPAGESIZE or a similar macro here?

microbit-carlos (commented):
Thanks Sam! I hope you don't mind I've added some minor notes.

martinwork:
This is good timing! I started work on the app today.

Could the python builder generate an 8 byte hash? The PF protocol transmits the hash from micro:bit to client in a struct with 8 bytes, to be checked against the hex. It looks like it's using hash type 2, which is a pointer to what looks like a long string, in which the first 8 bytes are always the same.

source/bluetooth/MicroBitPartialFlashingService.cpp
microbit-sam: It's so quick that I don't think that we need to worry about it for MakeCode, doesn't make a noticable difference when I do it
Perhaps I ought to verify it's actually doing the erasure..

I don't think backporting to the DAL is necessary, we don't ever intend to have BLE + MicroPython on V1 do we?

source/MicroBitMemoryMap.cpp
microbit-sam: Yeah - that's a good idea!

source/MicroBitMemoryMap.cpp
microbit-sam: 👍

microbit-sam:
> Could the python builder generate an 8 byte hash? The PF protocol transmits the hash from micro:bit to client in a struct with 8 bytes, to be checked against the hex. It looks like it's using hash type 2, which is a pointer to what looks like a long string, in which the first 8 bytes are always the same.

To update the hash for the filesystem region? Yes, I probably should update it although I'm not checking it anywhere at the moment

martinwork:
addlayouttable.py doesn't provide a hash for the filesystem region, and I don't check it either.

I'm referring to the mp/DAL region hash.

My understanding is that addlayouttable.py uses a pointer hash for the mp region, which is the equivalent of the DAL region for MakeCode, and is the hash we check, before a partial flash: https://github.com/microbit-foundation/micropython-microbit-v2/blob/master/src/addlayouttable.py#L211.

The mp/DAL region hash pointer points to microbit_version_string, which is initialised with MICROBIT_VERSION
https://github.com/microbit-foundation/micropython-microbit-v2/blob/a76e1413bcd66f128a31d98756fc3d1f336d1580/src/codal_port/modos.c#L36,
and is something like "micro:bit v2.0.0-beta.2+47ebc73-dirty on 2021-01-05; MicroPython v1.13 on 2021-01-05".

https://github.com/lancaster-university/codal-microbit-v2/blob/14f2a243ca869a56c10d5bb80e24c68d80290ab0/source/MicroBitMemoryMap.cpp#L176
`processRecord` copies 8 bytes from the string pointed to by the type 2 hash pointer. The first 8 bytes of that string "micro:bi" are no use. A unique version number or date & time might be OK, I suppose. Sending more that 8 bytes of hash will need a change to the protocol.

I have partial flashing working in the app with the existing PF service, by temporarily ignoring the hash, and instead checking the start and end addresses of the mp/dal region when the incoming hex is micropython. I'm using the filesystem region in the hex's table to decide what to send.

I measure (counting in my head!) a full flash at 66s and a partial flash as 24s, which feels very slow for a minimal hello world. Could the filesystem region (currently 6D000 to 73000) be made more compact and accurate, to just include the current script files?

microbit-sam:
Getting back to this

>  processRecord copies 8 bytes from the string pointed to by the type 2 hash pointer. The first 8 bytes of that string "micro:bi" are no use. A unique version number or date & time might be OK, I suppose. Sending more that 8 bytes of hash will need a change to the protocol.

I hadn't spotted this.. thanks Martin 

> I measure (counting in my head!) a full flash at 66s and a partial flash as 24s, which feels very slow for a minimal hello world. Could the filesystem region (currently 6D000 to 73000) be made more compact and accurate, to just include the current script files?

For partial flashing are you flashing the whole file system? I'm skipping blank pages after the script, as CODAL will erase the region from the last page written in flashing to the end of the region

martinwork:
> For partial flashing are you flashing the whole file system? I'm skipping blank pages after the script, as CODAL will erase the region from the last page written in flashing to the end of the region

Yes :( After finding the partial flash region, I was forgetting check what part of that was actually filled by the hex!

martinwork:
https://github.com/lancaster-university/codal-microbit-v2/blob/14f2a243ca869a56c10d5bb80e24c68d80290ab0/source/bluetooth/MicroBitPartialFlashingService.cpp#L325

It doesn't look like the calculated flashPointer will always be on a page boundary.


https://github.com/lancaster-university/codal-microbit-v2/blob/14f2a243ca869a56c10d5bb80e24c68d80290ab0/source/bluetooth/MicroBitPartialFlashingService.cpp#L344

This fixes up the settings pages after the PF in case it had a CRC validation, which the bootloader creates after a DFU. I overlooked the fact that we need to save no-validation settings pages before starting PF. Otherwise if PF fails the bootloader will drop into DFU mode. And if repeated partial flashings are like likely, it might be better to leave it as no-validation.

microbit-sam:
> It doesn't look like the calculated flashPointer will always be on a page boundary.

I've added a mask to find the page boundary

> This fixes up the settings pages after the PF in case it had a CRC validation, which the bootloader creates after a DFU. I overlooked the fact that we need to save no-validation settings pages before starting PF. Otherwise if PF fails the bootloader will drop into DFU mode. And if repeated partial flashings are like likely, it might be better to leave it as no-validation.

Ok good idea, so should the settings code be moved to where the KeyValue flag is set? Then a failed partial flash will enter pairing mode

I wonder if this should behave differently on Python, if the partial flash fails and MicroPython starts with a corrupt filesystem would the REPL still work..

martinwork:
> should the settings code be moved to where the KeyValue flag is set?

I reckon that's right. Get rid of the validation fix up at the end and ensure noValidation just before storing the flashIncomplete flag:
https://github.com/lancaster-university/codal-microbit-v2/blob/14f2a243ca869a56c10d5bb80e24c68d80290ab0/source/bluetooth/MicroBitPartialFlashingService.cpp#L292

That way only a failed full DFU flash will end up stuck in DFU mode.

```
void noValidation()
{
    // make bootloader settings page
    nrf_dfu_settings_t settings;
    memset( &settings, 0, sizeof( nrf_dfu_settings_t));
        
    // make NO_VALIDATION settings
    settings.settings_version = 2;
    settings.bank_0.bank_code = NRF_DFU_BANK_VALID_APP;
    settings.boot_validation_app.type = NO_VALIDATION;

    // make VALIDATE_CRC settings - calculate app size and CRC
    //settings.boot_validation_app.type = VALIDATE_CRC;
    //settings.bank_0.image_size = offset + 16 * sizeof( uint32_t) - MICROBIT_APP_REGION_START;
    //settings.bank_0.image_crc  = crc32_compute( (uint8_t*) MICROBIT_APP_REGION_START, settings.bank_0.image_size, NULL);
    //memcpy(settings.boot_validation_app.bytes, &settings.bank_0.image_crc, sizeof(uint32_t));

    // calculate settings page CRCs
    settings.crc = crc32_compute( (const uint8_t *)&settings.settings_version,
                                  offsetof(nrf_dfu_settings_t, init_command) - offsetof(nrf_dfu_settings_t, settings_version),
                                  NULL);

    settings.boot_validation_crc = crc32_compute( (const uint8_t *)&settings.boot_validation_softdevice, 3 * sizeof(boot_validation_t), NULL);

    uint32_t settingsSizeInWords = ( sizeof( nrf_dfu_settings_t) + sizeof(uint32_t) - 1) / sizeof(uint32_t);
        
    // save settings to flash settings page if different
    uint32_t *pSettings = (uint32_t *) MICROBIT_BOOTLOADER_SETTINGS;
    if ( memcmp( pSettings, &settings, sizeof( nrf_dfu_settings_t)))
    {
      flash.erase_page( pSettings);
      flash.flash_burn( pSettings, (uint32_t *)&settings, settingsSizeInWords);
    }
        
    // save settings to flash settings backup page if different
    uint32_t *pBackup   = (uint32_t *) MICROBIT_MBR_PARAMS;
    if ( memcmp( pBackup, &settings, sizeof( nrf_dfu_settings_t)))
    {
      flash.erase_page( pBackup);
      flash.flash_burn( pBackup, (uint32_t *)&settings, settingsSizeInWords);
    }
}

```

martinwork:
> I wonder if this should behave differently on Python, if the partial flash fails and MicroPython starts with a corrupt filesystem would the REPL still work.

I expected MicroPython would check the flashIncomplete flag, and boot to pairing mode to be consistent. If MicroPython can recognise a corrupt filesystem and not crash, I guess that would be OK, but only potentially useful if you can use the REPL. The no-validation settings would still be required, but maybe the flashIncomplete shouldn't be saved. What happens if the filesystem is changed via USB after a BLE full flash?

martinwork:
Sorry @microbit-sam. I think I might be wrong about the bootloader validation needing to be disabled before a partial flash. It's a while since I looked at it, but I have a feeling the validation only runs during the first reboot after a full DFU. The reason we invented a no validation setting was so that the application size didn't need to be customised when building a hex. No validation is interpreted as "size is not valid - do single bank dfu". So back to the way it was, I think, except putting it in a function is good.

martinwork:
@microbit-sam @microbit-carlos In a hello world python hex, the filesystem region has 256 bytes at the beginning (0x6d000), including some 0xFF bytes, and a lone byte at 0x72000. I guess a partial flash has to write pages 0x6d000 and 0x72000. Does it have to ensure the pages between, are erased?

microbit-carlos:
> and a lone byte at 0x72000.

Yes, this is the filesystem scratch page.

>Does it have to ensure the pages between, are erased?

Yes, otherwise if there were old files there they could be found by the filesystem.

martinwork:
So either the client app must send a block of 0xFFs for each empty page or the PF service needs to notice that pages are skipped and ensure they are blank.

microbit-carlos:
Yes, I think we agreed CODAL should notice it's the fs area and erase everything that is not 0xFF already. Is that the case @microbit-sam?

microbit-sam:
Yeah, that's what I'm aiming to do in this PR on line 382:
https://github.com/lancaster-university/codal-microbit-v2/pull/64/files#diff-aa2c3a0dceda8363e676a9836986800076f7bac72b3b9e99f354c8f09aba45deR382

Hopefully that makes sense and I've not messed it up!

martinwork:
I don't think the erasing code in the END_OF_TRANSMISSION case will do anything because of the byte in page 0x72000. The client will start at 0x6D000 and finish at 0x72000. The END_OF_TRANSMISSION code will start looking at 0x73000 and stop immediately. 

As I understand it, the pages in between will not be touched because the FLASH_DATA code erases a page only if it receives a block of packets with offset on the page boundary. 

At the moment, if the client has any data in a page, it must send something on the page boundary to trigger the erase, and it must send FFs for each empty page, if the middle pages have to be blank.

microbit-sam:
Hmm.. yeah that is annoying
Perhaps CODAL needs to be aware that when partial flashing micropython it must keep track of where it is before writing the layout table. A bigger change than I'd hoped for

Looking at the records that micropython currently uses, the layout table does not change between partial flashes
Regions 1 and 2 will remain in the same space with the same hashes, and the filesystem will stay in the same space and does not contain a hash

We could ignore the layout table in partial flashing and finish when data in the file system runs out, and erase from there until the layout table. This sounds like a bad idea but it would be simpler.. 

```
Python Layout Record. Record: 3 Hash Type: 0 Start: 0x6D000 Length: 0x6000 Hash: 0x0 0x0
Python Layout Record. Record: 2 Hash Type: 2 Start: 0x1C000 Length: 0x47664 Hash: 0x0 0x5B368
Python Layout Record. Record: 1 Hash Type: 0 Start: 0x1000 Length: 0x1B000 Hash: 0x0 0x0
```

martinwork:
> before writing the layout table.
Do you mean the "filesystem scratch page" 0x72000?
Yes, I think we want to avoid specialising the code for micro-python, but generalise is so it can cope with both.

We haven't dealt with sparse data before. It would be good to specify exactly what the app must send. At the moment, I 've modified the iOS app code to step through the region in 64 byte blocks (so not crossing a page boundary), sending a block (in 4 packets) if there are any non-FF bytes in it, or it's on a page boundary.

This seems like the simplest solution, except maybe FLASH_DATA should not erase the page if it's already blank.
https://github.com/lancaster-university/codal-microbit-v2/blob/50b8d901535147800ff5727ea5591e6f044d5b49/source/bluetooth/MicroBitPartialFlashingService.cpp#L347 

But I think that means the app is sending around double the number of packets it needs to for "hello world", in order to erase the blank pages.

Could the PF service remember the next page that will need to be erased (initially the first page in the region), then before each write in FLASH_DATA, erase all the non-blank pages from that "next" page up to the data's page?

And finish up by erasing from that "next" page up to the last page in the region? Could the micropython filesystem scratch page be empty sometimes? Does erasing to the end of the region make sense for MakeCode?

microbit-sam:
I've added a commit to erase pages between last page written + 1, and the current page
However I think this would fall over if the packets were out of order

> This seems like the simplest solution, except maybe FLASH_DATA should not erase the page if it's already blank.

Yep I've got a CRC in there to check for blank pages

microbit-sam:
The latest commit creates a table of pages, and marks which pages have been touched
`pagesTouched = (uint8_t *)malloc(MICROBIT_CODESIZE); `

And then erases untouched pages when partial flashing completes

```
      // Erase untouched pages in filesystem
      int start = (FLASH_PROGRAM_END / MICROBIT_CODEPAGESIZE);
      int end   = (MICROBIT_APP_REGION_END / MICROBIT_CODEPAGESIZE);
      for(int page = start; page < end; page++) {
        DMESG("Checking page: %u", page);
        if(!pagesTouched[page]) {
            DMESG("Erasing page: %u", page);
            flash.erase_page( (uint32_t *) (page * MICROBIT_CODEPAGESIZE));
        }
      }
```

Maybe pagesTouched should only be (MICROBIT_APP_REGION_END - FLASH_PROGRAM_END) long

martinwork:
I have a horrible feeling we have invented a really complicated way to erase all the pages from the start of the PF region to the end of the app region! If all pages that do not contain data must be blank, and all pages that do contain data must be blank in order to burn the data, isn't that all the pages?

The dal/codal code in master only changes a "no-data" page (without an erase) if it finds the MakeCode embedded script magic. Why was that necessary? Why is erasing all pages now necessary?  Will this change the use of MicroBitFileSystem?

The check for "already blank" has gone:
https://github.com/lancaster-university/codal-microbit-v2/blob/944e74693f29c0424caec191d7d1f9da9190d1c7/source/bluetooth/MicroBitPartialFlashingService.cpp#L353
https://github.com/lancaster-university/codal-microbit-v2/blob/944e74693f29c0424caec191d7d1f9da9190d1c7/source/bluetooth/MicroBitPartialFlashingService.cpp#L394

Depending on the value of blockNum, I think the following might burn the same data again or some junk or to a page that hasn't been erased yet, or to an invalid offset.
https://github.com/lancaster-university/codal-microbit-v2/blob/944e74693f29c0424caec191d7d1f9da9190d1c7/source/bluetooth/MicroBitPartialFlashingService.cpp#L378

> Maybe pagesTouched should only be (MICROBIT_APP_REGION_END - FLASH_PROGRAM_END) long

Yes, or the number of pages in the PF region. A micropython PF only potentially touches 6 pages out of 128, but it would be bigger for a MakeCode app.

>       int start = (FLASH_PROGRAM_END / MICROBIT_CODEPAGESIZE);
>       int end   = (MICROBIT_APP_REGION_END / MICROBIT_CODEPAGESIZE);
https://github.com/lancaster-university/codal-microbit-v2/blob/944e74693f29c0424caec191d7d1f9da9190d1c7/source/bluetooth/MicroBitPartialFlashingService.cpp#L388
Shouldn't this be working on the PF region? Could this erase the last page of the application code and the micropython region table?
The pagesTouched array is only needed during partial flashing in pairing mode. Is it safe that it's only initialised in the constructor? The amount of heap available in pairing mode doesn't depend on the app, so wouldn't it be OK to lazily allocate and initialise pagesTouched, block and the other PF-only variables when needed? Then it wouldn't matter that pagesTouched was unnecessarily large.

microbit-sam:
> isn't that all the pages?

*facepalm* Yes, of course

> The dal/codal code in master only changes a "no-data" page (without an erase) if it finds the MakeCode embedded script magic. Why was that necessary? Why is erasing all pages now necessary? Will this change the use of MicroBitFileSystem?

Previously when the MakeCode source was embedded in the hex file it would invalidate the source so that if the hex was extracted, it would no longer load into MakeCode. We don't need this in CODAL

So best method would be to erase the filesystem at the start of partial flashing, then write data wherever it needs to go, and not have to worry about erasing pages during/after

Looking at MicroBitFileSystem it currently has a format function, that doesn't erase pages. Adding a parameter to also erase makes sense to me?
https://github.com/lancaster-university/codal-microbit-v2/blob/de856f22f7fbea1145c24fff7933dd12fc30b2ae/source/MicroBitFileSystem.cpp#L250

martinwork:
I'm not sure what you mean about MicroBitFileSystem. I was just wondering if it was generally preserved through a MakeCode partial flash before, but erasing all pages would rule that out and increase wear on the flash. Currently, a MakeCode partial flash doesn't erase any pages after the PF.

microbit-sam:
> I'm not sure what you mean about MicroBitFileSystem. I was just wondering if it was generally preserved through a MakeCode partial flash before, but erasing all pages would rule that out and increase wear on the flash. Currently, a MakeCode partial flash doesn't erase any pages after the PF.

I'm thinking in terms of Python. The Python script(s) that execute are stored in the filesystem, including the main.py that the program enters to start. So for a Python partial flash I think we have to erase the whole filesystem to ensure there's no scripts left

Erasing the filesystem using MicroBitFileSystem rather than manually erasing pages would also mean that if the filesystem (or part of it) was moved to the KL27 in the future it would be handled correctly already

For MakeCode, I'm not sure whether erasing the filesystem is the right thing to do. It would be cool if you could still access any data you may have logged, but also it might be completely irrelevant!

microbit-sam:
I've changed it to reformat the filesystem when partially flashing

jaustin:
@microbit-sam does that mean you think this is ready for being merged then?

microbit-sam:
I think so, but as the previous comment thread demonstrates there may be something I've not considered

One worry I had was that erasing all of the pages in the filesystem might take so long that it causes an issue with the first packet. This doesn't seem to be the case when I've been using it (I ran it forcing all pages to be erased with no problems), but if the filesystem was larger using the KL27's free space maybe it'd be slower and cause problems

martinwork:
I still don't understand the need to erase all the trailing pages for a MakeCode PF. They didn't need to be erased before.

I didn't understand the link with MicroBitFileSystem or moving it to the KL27. Will MicroPython's file system move there too? In any case MicroPython doesn't use MicroBitFileSystem, does it?

MicroBitFileSystem:reformat() formats the filesystem, which I think means that a page or two will no longer be blank, and need to be erased again to PF. Isn't it in danger of deleting the MicroPython memory map table? I don't think MicroBitFileSystem has to have all pages erased to be reinitialised.

Is all-FFs the only combination of bytes that has that CRC?

microbit-sam:
> I still don't understand the need to erase all the trailing pages for a MakeCode PF. They didn't need to be erased before.

Previously we only had to remove the embedded source, and now we don't. I think I've removed all the trailing MakeCode erasing code now (apart from the filesystem reformat)

> I didn't understand the link with MicroBitFileSystem or moving it to the KL27. Will MicroPython's file system move there too? In any case MicroPython doesn't use MicroBitFileSystem, does it?

MicroPython stores it's scripts and executes from the file system
https://github.com/microbit-foundation/micropython-microbit-v2/blob/a76e1413bcd66f128a31d98756fc3d1f336d1580/src/codal_port/main.c#L49

> MicroBitFileSystem:reformat() formats the filesystem, which I think means that a page or two will no longer be blank, and need to be erased again to PF. Isn't it in danger of deleting the MicroPython memory map table? I don't think MicroBitFileSystem has to have all pages erased to be reinitialised.

Erasing the filesystem should preserve the layout table
```
SoftDevice    0x01000..0x1c000
MicroPython   0x1c000..0x63a58
Layout table  0x63fc0..0x64000
Filesystem    0x6d000..0x73000
```

`MBFS::init` has built in checks to prevent it from overwriting an existing filesystem
and `int MicroBitFileSystem::format()` states
`* Initialises a new file system. Assumes all pages are already erased.`

I thought it made sense to create a seperate function to erase and then call `format` rather the modify the existing parts. But maybe for partial flashing reformating isn't necessary (as you said, the first page will be erased almost instantly), and just having an erase all function to clear the filesystem is better

> Is all-FFs the only combination of bytes that has that CRC?

It's a good question, and I'm not certain how to verify. Searching the internet for the hex string gave me two results, one a forum thread about blank pages, and one some source from VMWare also searching for blank pages

However this has made me realise that `0xc71c0011` is for an all 0s page, not an all 0xFF page and will need to be changed

martinwork:
> Previously we only had to remove the embedded source, and now we don't.

So why erase all those pages that didn't need to be touched before?

> MicroPython stores it's scripts and executes from the file system

I think micropython has it's own file system and does not use MicroBitFileSystem.

https://github.com/microbit-foundation/micropython-microbit-v2/blob/a76e1413bcd66f128a31d98756fc3d1f336d1580/src/codal_port/main.c#L56
https://github.com/microbit-foundation/micropython-microbit-v2/blob/a76e1413bcd66f128a31d98756fc3d1f336d1580/src/codal_port/main.c#L39
https://github.com/microbit-foundation/micropython-microbit-v2/blob/master/.gitmodules#L3
https://github.com/micropython/micropython/blob/master/ports/nrf/modules/uos/microbitfs.c#L180

> Erasing the filesystem should preserve the layout table

The micropython table is at the end of the page containing FLASH_PROGRAM_END/mp_end **or the next page** if there's no room in that page.
https://github.com/microbit-foundation/micropython-microbit-v2/blob/master/src/addlayouttable.py#L202
https://github.com/microbit-foundation/micropython-microbit-v2/blob/master/src/addlayouttable.py#L218

In the example, the layout table happens to be in the same page as FLASH_PROGRAM_END. MicroBitFileSystem::reformat starts erasing at  the first page boundary at or after FLASH_PROGRAM_END, which could contain the layout table in a future micropython build.
https://github.com/lancaster-university/codal-microbit-v2/pull/64/files#diff-9d1778509c3b5f7217f70a176bd7c06c1b756893999ad3d56ab38f8bf71dd303R251

In any case, reformat() calls format(), which saves data in at least one page, so it's no longer blank.

I think it would also be wise to avoid trampling on the space between the layout table and the micropython filesystem. The layout table specifies the address range of the micropython filesystem, which is the area the PF code should work on.

MicroBitFileSystem::reformat assumes too much about the filesystem start address. MakeCode starts the MicroBitFileSystem at a different address.

I think having the PF code use MicroBitFileSystem is unnecessarily complicating the situation. Assuming that a PF should erase the MicroBitFileSystem is especially problematic if MicroBitFileSystem were to move into KL27 memory. 

> and int MicroBitFileSystem::format() states
> * Initialises a new file system. Assumes all pages are already erased.

I hadn't noticed that. I think it isn't true. Maybe it was in a early version?

> However this has made me realise that 0xc71c0011 is for an all 0s page, not an all 0xFF page and will need to be changed

This didn't occur to me at first, but there are 4,294,967,296 possible checksums and 256^4096 combinations of bytes in a page. That's a very large number of combinations per checksum (>3e+9854). A simple scan for any non-FF would be safer and quicker.

microbit-sam:
> I think micropython has it's own file system and does not use MicroBitFileSystem.

Oh sorry, you're right. I've always thought they were one and the same. I'll rework this
Thanks Martin

> So why erase all those pages that didn't need to be touched before?

Is this the MBFS pages that I'm erasing (incorrectly, I've now realised)? I'll remove that and look at clearing the MicroPython filesystem now

If you mean the embedded source page, it should have previously only removed the embedded source magic, and nothing else. Can you link to the line you're concerned about as I'm not sure where you mean

> The micropython table is at the end of the page containing FLASH_PROGRAM_END/mp_end or the next page if there's no room in that page.

I think once I change to erasing the correct filesystem, I'll remove the changes to MicroBitFilesystem to keep this PR tidy, and I'll bare this in mind when doing so

> A simple scan for any non-FF would be safer and quicker.

True, I guess the CRC will have to iterate through each byte anyway so I might as well check them myself

martinwork:
I'm sorry I mentioned the MicroBitFileSystem! I had a feeling they were different, but hadn't tracked down the micropython code until today. 

> If you mean the embedded source page, it should have previously only removed the embedded source magic, and nothing else.
Yes, in the current master, this code overwrites the MakeCode source magic. Because it writes zero, it will not erase the page.
https://github.com/lancaster-university/codal-microbit-v2/blob/master/source/bluetooth/MicroBitPartialFlashingService.cpp#L333

Early on in this PR, I think this changed to erasing all pages after the PF data, which is required for MicroPython, short of parsing the MicroPython file system. After a quick look at the MicroPython file system code, it appears the page at the end starting FD will never be blank, because that's a page type byte, but that page doesn't have to be at the end of the region.

My instinct is to try not to, to avoid side effects, but it may be that erasing the PF area is the best simple solution, There are a lot of cases to think about, and it only potentially means one extra erase per page (and wiping the MicroBitFileSystem). 

There's an instant in a MakeCode PF where the MakeCode magic is erased. That instant will be a while if all pages are erased before starting to write. 

MicroBitMemoryMap can tell, but doesn't record or return if it's MakeCode or MicroPython.

source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -289,16 +349,40 @@ void MicroBitPartialFlashingService::partialFlashingEvent(MicroBitEvent e)
        
        KeyValuePair* flashIncomplete = storage.get("flashIncomplete");
        if(flashIncomplete == NULL){
+
          uint8_t flashIncompleteVal = 0x01;
          storage.put("flashIncomplete", &flashIncompleteVal, sizeof(flashIncompleteVal));
+            
+         // Check if FS exists
+         if(micropython_fs_end != 0x00) {
+            for(uint32_t *page = (uint32_t *)micropython_fs_start; page < (uint32_t *)(micropython_fs_end - MICROBIT_CODEPAGESIZE); page += (MICROBIT_CODEPAGESIZE / sizeof(uint32_t))) {
```
microbit-sam, line 358: I'm not sure why but I'm having to skip the final page of the filesystem
After partial flashing the script won't run. Looking at the page before/after flashing there's a single byte different at `0x72000`

@microbit-carlos is that an important flag for MicroPython? I'm not sure why it's there to start off with, but it seems important!

```
diff pre.hex post.hex
2c2
< :10200000FDFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFE2
---
> :10200000FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFE0
```

carlosperate: That byte is the indicator of the "persistent page" (a scratch page for the filesystem): https://github.com/micropython/micropython/blob/b0932fcf2e2f9a81abf7737ed4b2573bd9ad4a49/ports/nrf/modules/uos/microbitfs.c#L79

With the microbit-fs library we place it as the last page in flash, which would be 0x7200 as right now the fs ends in 0x73000 https://github.com/microbit-foundation/micropython-microbit-v2/blob/master/src/codal_port/filesystem.ld#L3

microbit-sam: Ah yeah, I remember now

So is it best for partial flashing to erase pages up to: 'end of fs' - 1 page, to avoid unecessary erases

Partial flashing should have put the 0xFD in regardless, so I think there's a bug there which I'll have to look at anyway

microbit-carlos: This page could actually be placed at the top or bottom of the filesystem flash area, so the ` 'end of fs' - 1` rule is not a guarantee.
Nevertheless, as you mentioned, the hex file from a Python Editor should include that byte anyway, so partial flashing should detect that page as being the same and not touch it?

microbit-sam: Ok cool, I'll make sure to erase the full region then
Yeah, it should be flashed so I'll check whats up with that

microbit-sam:
My hex parser was ignoring the scratch page, fixed that and this works as expected

microbit-sam:
Hey @martinwork, please could you give this a quick look

source/MicroBitMemoryMap.cpp
```diff
@@ -138,5 +142,64 @@ void MicroBitMemoryMap::findHashes()
...
+        if (   *(magicAddress - (sizeof(magicAddress))) == 0x597F30FE
+            && *(magicAddress - (sizeof(magicAddress)/4))   == 0xC1B1D79D
+           ) {
+            // Found uPy Magic
+            DMESG("FOUND UPY_MAGIC");
+            uint8_t nRegions = (*(magicAddress - (sizeof(magicAddress)/2)));
+            
+            // Iterate through regions
+            for(int i = 0; i < nRegions; i++) {
+               processRecord(magicAddress - ((2 + i) * sizeof(magicAddress)));
```
martinwork, line 160: I think this will break if another region was added to the table.

source/MicroBitMemoryMap.cpp
```diff
@@ -138,5 +142,64 @@ void MicroBitMemoryMap::findHashes()
...
+    if(record.ht == HASH_TYPE_POINTER) {
+        // Points to a string
+        // Take string and generate 8 byte hash
+        uint8_t * string = (uint8_t *) record.hash_data[0];
+        int n   = 0;
+        while(n < MAX_STRING_LENGTH && *(string + n) != '\0') {
+            n++;
+        }
+
+        uint32_t crc = crc32_compute(string, n, NULL);
```
martinwork, line 198: Is a 4 byte CRC good enough for this?

source/bluetooth/MicroBitPartialFlashingService.cpp
```diff
@@ -289,16 +338,29 @@ void MicroBitPartialFlashingService::partialFlashingEvent(MicroBitEvent e)
...
        }
        delete flashIncomplete;
 
       uint32_t *flashPointer   = (uint32_t *)(offset);
 
       // If the pointer is on a page boundary erase the page
-      if(!((uint32_t)flashPointer % MICROBIT_CODEPAGESIZE))
-          flash.erase_page(flashPointer);
+      if(!((uint32_t)flashPointer % MICROBIT_CODEPAGESIZE)) {
+          if(crc32_compute((uint8_t*)flashPointer, MICROBIT_CODEPAGESIZE, NULL) != 0xF154670A) {
```
martinwork: Checking the page for any non-FFs would be more robust.

martinwork:
I have added a few comments.

source/bluetooth/MicroBitPartialFlashingService.cpp
microbit-sam: In the latest commit it is checking for non-FF

source/MicroBitMemoryMap.cpp
microbit-sam: The string is :
`micro:bit v2.0.0-beta.1+130a254-dirty on 2020-12-15; MicroPython v1.13 on 2020-12-15`

So we'd have to pick the 4 byte git hash from the string, (or could event extract it and use it?)

source/MicroBitMemoryMap.cpp
microbit-sam: Yeah you're right as the struct has a set size. I'm not sure what the best way to avoid this is, if we allocate more now or later when we need it

From `inc/MicroBitMemoryMap.h`
```
#define NUMBER_OF_REGIONS 3
    struct MemoryMapStore
    {
        Region memoryMap[NUMBER_OF_REGIONS];
    };
```

martinwork: Oh! I didn't spot that you are using nRegions. I was just thinking the code uses the last few regions in the table. I had better check my code! I think it will be OK if processRecord ignores any new record.id's that don't make an index that fits in the array.

microbit-sam: 👍

source/MicroBitMemoryMap.cpp
microbit-sam: Oh, do you mean the resulting CRC from the operation?
I think we have space for 8 byte hashes, so it could halve the string and generate a hash for each half

martinwork: I'm not saying the CRC isn't good enough. A false match probably wouldn't brick micro:bit. I think it would be more sensible if the python tool that creates the table made eight bytes that identify the micropython-microbit-v2 build. Maybe the UTC of the build would be OK for most uses, though inconvenient for testing.

The string is
```
"micro:bit v" MICROBIT_RELEASE "+" MICROBIT_GIT_HASH " on " MICROBIT_BUILD_DATE \
    "; MicroPython " MICROPY_GIT_TAG " on " MICROPY_BUILD_DATE
``` 

I'm not sure where MICROBIT_GIT_HASH and MICROPY_GIT_TAG come from, or what format we could rely on.

source/bluetooth/MicroBitPartialFlashingService.cpp
martinwork: Sorry! I must have got hooked into the wrong version. This looks good now.

source/MicroBitMemoryMap.cpp
microbit-sam: So along the lines of?
It will defo avoid the collision issue, but you're right it will be annoying for testing!
```
--- a/src/addlayouttable.py
+++ b/src/addlayouttable.py
@@ -46,6 +46,7 @@ import argparse
 import binascii
 import struct
 import sys
+import time

 IHEX_TYPE_DATA = 0
 IHEX_TYPE_EXT_LIN_ADDR = 4
@@ -209,7 +210,7 @@ def main():
     layout = FlashLayout()
     layout.add_region(1, sd_start, sd_end - sd_start, FlashLayout.REGION_HASH_NONE)
     layout.add_region(
-        2, mp_start, mp_end - mp_start, FlashLayout.REGION_HASH_PTR, mp_version
+        2, mp_start, mp_end - mp_start, FlashLayout.REGION_HASH_DATA, hex(int(time.time()))
     )
     layout.add_region(3, fs_start, fs_end - fs_start, FlashLayout.REGION_HASH_NONE)
     layout.finalise()
```

martinwork: It was just an idea. I don't know what hex(int(time.time()) produces, or exactly what time.time() does for that matter! As released versions wouldn't be generated too rapidly, maybe one second precision is not required.

MICROBIT_BUILD_DATE and MICROPY_BUILD_DATE seem irrelevant as to whether a full flash is required.

I suppose MICROBIT_GIT_HASH and MICROPY_GIT_TAG aren't available to python? And I can't see how to make 8 bytes from them.

I don't think we've ever had a problem with the hash MakeCode uses, but it ought to be possible to make a simple version number for the python hex.

microbit-sam: I spoke to Carlos and Jonny and they would prefer to keep the pointer, it provides easy access to the version string for any program that can parse the hex

So if we keep the pointer and version string, should we make this 8 bytes by splitting it in two? Or using a 64bit hash? Or keep it simple and as it is with the CRC32

martinwork: Maybe keep it simple and see if we notice a problem?

I would change
```
        uint32_t crc = crc32_compute(string, n, NULL);
        memcpy(&hash, &crc, 4);

```
to 
```
        hash[0] = crc32_compute(string, n, NULL);
        hash[1] = 0;

```

martinwork: How do you calculate the CRC in the Android app?

finneyj:
thanks @microbit-sam  @martinwork !

PR #71, opened 2021-01-28
Fixing Inaccurate notes produced on Micro:bit v2

JoshuaAHill:
Changed variable sampleRate to float to avoid rounding errors when generating frequencies so that tones played via the microbit v2 internal speaker are now more accurate to the target frequency, during testing only 0.1-0.4hz out, which could be due to innacuracy in the detection equipment.
Additionally, manually set the mixer2 sample rate to 44100 in the microbit audio enable function. Some experinentation could be done here to match the sample rate to the true output from the PWM and observe if this improves accuracy futher.

This pull request mirrors and relies on lancaster-university/codal-nrf52 repo of the same name


finneyj:
Thanks @JoshuaAHill - looks great. The internal samplerate information was always held as a float anyway, so this makes perfect sense.

PR #75, opened 2021-02-01
MicroBitButton: pin becomes pin.name

microbit-sam:
Thanks for spotting @martinwork


finneyj:
Thanks @microbit-sam @martinwork.

Looks good.

PR #103, opened 2021-05-12
Deep sleep 2

martinwork:
Sibling PRs in codal-core, codal-nrf52, codal-microbit-v2


source/MicroBitPowerManager.cpp
```diff
@@ -370,40 +427,217 @@ void MicroBitPowerManager::deepSleep(uint32_t milliSeconds)
...
-    // Enable an interrupt on the requested pin or an interrupt from the KL27.
-    pin.setDetect(pin.getPolarity() ? GPIO_PIN_CNF_SENSE_High : GPIO_PIN_CNF_SENSE_Low);
+            // Enable an interrupt on the requested pin or an interrupt from the KL27.
+            wakeUpPin->setDetect(wakeUpPin->getPolarity() ? GPIO_PIN_CNF_SENSE_High : GPIO_PIN_CNF_SENSE_Low);
+        }
+    }
 
     // Enable wakeup from the the KL27 interrupt line.
     io.irq1.setDetect(GPIO_PIN_CNF_SENSE_Low);
+    NVIC_EnableIRQ(GPIOTE_IRQn);
```
finneyj, line 546: hi @martinwork 

GPIOTE is used in lots of places... Why do we need this? I'm a bit concerned there may be a lot of IRQs from light sensor, display....

martinwork: While not at all sure, I thought I confirmed that GPIOTE_IRQn seems to be necessary to get wake up IRQs from pins and buttons. This line is a bit of a hang over from a previous attempt which disabled then enabled it on the way in and out, but I thought as it's needed for pins, it would be needed for io.irq1. How do I send irq1 low?

martinwork: If it is required, all those places will have to disable their use of  GPIOTE_IRQn, and if it's desirable to disable GPIOTE_IRQn when it's not needed we'll have to ask them all if they need it. I believe deepSleep(MicroBitPin &pin) was running with GPIOTE_IRQn enabled.

source/bluetooth/MicroBitBLEManager.cpp
```diff
@@ -1530,4 +1569,87 @@ static void microbit_dfu_evt_handler(ble_dfu_buttonless_evt_type_t event)
...
 #endif // CONFIG_ENABLED(MICROBIT_BLE_DFU_SERVICE)
 
+
+#define MICROBIT_PANIC_NRF_BASE 70
+#define MICROBIT_PANIC_NRF_FAULT_ID_SD_ASSERT  (MICROBIT_PANIC_NRF_BASE)
+#define MICROBIT_PANIC_NRF_FAULT_ID_APP_MEMACC (MICROBIT_PANIC_NRF_BASE + 1)
+#define MICROBIT_PANIC_NRF_FAULT_ID_SDK_ASSERT (MICROBIT_PANIC_NRF_BASE + 2)
+#define MICROBIT_PANIC_NRF_FAULT_ID_SDK_ERROR  (MICROBIT_PANIC_NRF_BASE + 3)
+#define MICROBIT_PANIC_NRF_FAULT_UNKNOWN       (MICROBIT_PANIC_NRF_BASE + 4)
+
```
martinwork, line 1579: @finneyj Are these panic number assignments OK?

finneyj: 07x is currently free:
https://github.com/lancaster-university/codal-core/blob/master/inc/core/ErrorNo.h#L80

I don't think we can really reserve a code for NRF though - too specific. We could reserve 07x as CPU/SDK specific codeset though. It would make sense to have one of those.

martinwork: So add DEVICE_CPU_SDK = 70 to ErrorNo and use that instead of MICROBIT_PANIC_NRF_BASE?

finneyj: i think keeping NRFxxx is fine here (target repo). But we could reserve 07x in codal-core/inc/errno.h to be CPU specific codes

finneyj:
Thanks for all the great work here @martinwork! This is a great feature for v2.

I've done a code review encompassing all the three PRs across codal-core, codal-nrf52 and codal-microbit-v2. generally, I think we're getting super close to be able to merge this now. I'm pasting the review in here, just for tracking really, so we don't get split across three repos... thoughts and discussion welcomed of course!

Deep Sleep 2 Code Review:

This is a big PR, spanning three repos (codal-core, codal-nrf52, codal-micrbit-v2).
This is however in the nature of power management - it spans many subsystems.
Generally, I feel this has been thoughtfully and considerately done. Good stuff.

There are a few architectural things I would have preferred:
  - The changes to codal-core and codal-nrf52 have been minimized. 
  - I can see why you chose to do this (reduce surface area of change), however this does lead
    somewhat to a breakdown in encapsualtion in those repos, and a lack of reuse. It also leads
    to lots of low level code in the codal-microbit-v2 power manager. Which is ok, but not great.
  - It would dhave been better to set up a driver-model for the power manager in codal-core, and
    migrate as much as possible into the codal-core and codal-nrf52 repos.

Having said that, such code can be achieved further down the line, and I don't think this holds up
merging this PR for next release, as this is a great feature to have.

Here's a list of other minor things I thought of while reviewing the code. Would be good if you
can consider these before raising final PR for merge. Again, we can patch up later if needed, but
would be good to get as many fixed as possible pre-merge:


  - There's a tendency to use pointer in/out parameters in the code. 
     e.g. Timer::deepSleepBegin
          Timer::deepSleepWakeUpTime
          CodalComponent::manageSleep()

  - I've tried to avoid that in CODAL as it can be a bit hard to wrap in highler level languages
    sometimes. Also, can make some calls a bit inconsistent (e.g. note deepSleepBegin vs deepSleepEnd).
    Minimize this pattern if you can, but not at the risk of major changes.

  - NRF52ADC: Use flag rather than static vars for holding state here? In case we have >1 
    ADC in future...

  - getWakeOnActive() should really be isWakeOnActive()

  - We now seem to have two levels of component callback function for dealing with DeepSleep. 
    We now have both setSleep() with manageSleep(). i can see why you needed more fine
    grained control, but would have been better to just power-up one implementaiton rather than
    wrap it. The name manageSleep() is also not great. Renaming as deepSleepCallback() to align
    with idleCallback() etc would have been better. Then just use that consistently... 
    Again though, this could be cleaned up later.

  - Quite a few of the doxygen(javadoc style) comments don't really deacribe what the method DOES.
    Often only describing when the method is called. Would be good to update these with a summary,
    and full description of parameters please.

  - Still quite a few inline comments that are still WIP, DMESG etc. Please tidy those up.

  - Changes to LSM303 look ok, but remember we dual footprint FXOS8700 on v2 also. Does that driver
    need equivalent treatment?
  
  - NRF52Serial::setSleep() can fail and return SERIAL_IN_USE... what happens here?  

  - MicroBitIO seems to reuse value '1' as a flag : IO_SAVED_STATUS_SAVED. Is this safe?

  - MicroBitPowerManager. No need for a new constructor here to maintain transparency. No need to
    overload - just replace the old one.

  - MICROBIT_POWER_MANAGER_MINIMUM_DEEP_SLEEP/MICROBIT_POWER_MANAGER_MINIMUM_POWERUP. 
  - Can we rename as CONFIG options?
  - e.g. CONFIG_MINIMUM_DEEP_SLEEP_TIME and CONFIG_MINIMUM_POWER_ON_TIME

  - clearWakeUpSources(). What is this for? and why does it not clear timer events? 
    Seems a bit... inconsistent/partial? Do we need this?

  - {manageSleepBeginWithWakeUps,manageSleepEndWithWakeUps} / {manageSleepBegin,manageSleepEnd}
    can these be combined? The callback function feel fiercly overloaded?
  
  - deepSleep(ms): is this the correct behaviour for MIN? Should we simply NOT DEEPSLEEP at ALL 
    if requested time is < MIN? rather than round it up to MIN?

  - deepSleep(ms): are calls to fiber_sleep() and cancel_event() in the right order here? 
    Should we cancel before sleeping?

  - app_error_fault_handler(). Do we really want DMESGF here?? could cause hangs if happens 
    in IRQ context, before panic()...

martinwork:
Thanks @finneyj ! Yes, my thinking was to limit changes below codal-microbit-v2, but ended up needing more changes than I expected. Probably best to leave that as-is for now. Thanks for collecting a list of issues. I'll start working through them today.

martinwork:
NRF52ADC: Use flag rather than static vars for holding state here? In case we have >1
ADC in future...

Changes to LSM303 look ok, but remember we dual footprint FXOS8700 on v2 also. Does that driver
need equivalent treatment?

NRF52Serial::setSleep() can fail and return SERIAL_IN_USE... what happens here?

MicroBitIO seems to reuse value '1' as a flag : IO_SAVED_STATUS_SAVED. Is this safe?

{manageSleepBeginWithWakeUps,manageSleepEndWithWakeUps} / {manageSleepBegin,manageSleepEnd}
can these be combined? The callback function feel fiercly overloaded?

deepSleep(ms): is this the correct behaviour for MIN? Should we simply NOT DEEPSLEEP at ALL
if requested time is < MIN? rather than round it up to MIN?

deepSleep(ms): are calls to fiber_sleep() and cancel_event() in the right order here?
Should we cancel before sleeping?

Quite a few of the doxygen(javadoc style) comments don't really deacribe what the method DOES.
Often only describing when the method is called. Would be good to update these with a summary,
and full description of parameters please. ​

**Partially done**

> We now seem to have two levels of component callback function for dealing with DeepSleep.
> We now have both setSleep() with manageSleep(). i can see why you needed more fine
> grained control, but would have been better to just power-up one implementaiton rather than
> wrap it. The name manageSleep() is also not great. Renaming as deepSleepCallback() to align
> with idleCallback() etc would have been better. Then just use that consistently...
> Again though, this could be cleaned up later.
> 
I've changed the names to deepSleepCallback() and deepSleepAll().
I wasn't sure I could replace the existing setSleep API (should have asked!) and changing all the ::setSleep overrides which didn't need the extra control, seemed like a lot of churn.

> Quite a few of the doxygen(javadoc style) comments don't really deacribe what the method DOES.
> Often only describing when the method is called. Would be good to update these with a summary,
> and full description of parameters please.
> 
Done for Timer

> There's a tendency to use pointer in/out parameters in the code.
> e.g. Timer::deepSleepBegin
> Timer::deepSleepWakeUpTime
> CodalComponent::manageSleep()
> 
Done for Timer. Leaving manageSleep for now, because most uses send a NULL pointer.

**Done**

> getWakeOnActive() should really be isWakeOnActive()

> Still quite a few inline comments that are still WIP, DMESG etc. Please tidy those up.
I think I got 'em.

> MicroBitPowerManager. No need for a new constructor here to maintain transparency. No need to
> overload - just replace the old one.
> 
Deleted the old one.

> MICROBIT_POWER_MANAGER_MINIMUM_DEEP_SLEEP/MICROBIT_POWER_MANAGER_MINIMUM_POWERUP.
> Can we rename as CONFIG options?
> e.g. CONFIG_MINIMUM_DEEP_SLEEP_TIME and CONFIG_MINIMUM_POWER_ON_TIME
> 

clearWakeUpSources(). What is this for? and why does it not clear timer events?
Seems a bit... inconsistent/partial? Do we need this?

Deleted it. 

> app_error_fault_handler(). Do we really want DMESGF here?? could cause hangs if happens
> in IRQ context, before panic()...
> 
Replaced with DMESG, so they won't get in the way of the panic. I guess a DMESGF will be needed to see them when debugging.

martinwork:
Please check @finneyj 

> deepSleep(ms): is this the correct behaviour for MIN? Should we simply NOT DEEPSLEEP at ALL
> if requested time is < MIN? rather than round it up to MIN?
> 
Good point. I suppose I was thinking... to sleep less than MIN, use fiber_sleep().
I've changed the behaviour to match the following. Does that make more sense?
```
If the requested time interval is less than CONFIG_MINIMUM_DEEP_SLEEP_TIME, or a wake-up timer
cannot be allocated, the sleep will be a simple uBit.sleep() without powering down.
```
Needs testing again.

> deepSleep(ms): are calls to fiber_sleep() and cancel_event() in the right order here?
> Should we cancel before sleeping?
> 
I expect I had it the other way round at some point, but I think it's right as it is. When some other IRQ triggers a power up, it comes out of deepSleepWait and enters fiber_sleep(). It doesn't powerDownDisable(), so the code that processes that other reason for waking up can cause another power down. Then it needs its wake-up timer still in place, to trigger another power up when the original time interval expires.

finneyj:
Thanks @martinwork - they all sound like sensible improvements!

finneyj:
Thanks for all the hard work @martinwork.

PR #104, opened 2021-05-13
Codal fs

finneyj:
Introduction of persistent logging to USB interface.

  - enables ~120KB of persistent storage for user use
  - introduces CSV logging capability, with integrated HTML preview and download on MICROBIT drive


jamesadevine:
Killer! Nice work all involved!

Issue #108, opened 2021-05-18
MicroBitLog uses innerHTML wich leads to XSS attacks


pelikhan:
The user is able to inject HTML/Javascript into the data page.

```
input.onButtonPressed(Button.AB, function() {
    flashlog.clear()
})
flashlog.setTimeStamp(FlashLogTimeStampFormat.Milliseconds)
basic.forever(function () {
    led.toggle(0, 0)
    const ax = input.acceleration(Dimension.X)
	flashlog.beginRow()
    flashlog.logData(`a.x`, `<a href="https://bing.com">${ax}</a>`)
    flashlog.endRow()
})
```

pelikhan:
Recommendation is to use DOM operation to set the value of cells instead of innerHTML.

microbit-matt-hillsdon:
This issue can be fixed by using innerText or other DOM API rather than innerHTML, but there's a related injection issue due to the use of a comment to package the data.

There's nothing preventing user text terminating the comment, outputting arbitrary HTML/JS (and potentially resuming the comment).

E.g. (in C++ syntax):

```
uBit.log.logData("xss2", "--><svg onload='alert(2)'><!--");
```

I think this will have to be addressed on the C++ side.

https://html.spec.whatwg.org/multipage/syntax.html#comments

finneyj:
Thanks @microbit-matt-hillsdon 

yep - thanks for the reminder! I always meant to prevent the user from adding `-->` into the datastream...

If you can patch up my dodgy HTML @microbit-matt-hillsdon  I'll just filter the `-->` symbol out at ingress?

pelikhan:
Why are you using a text format? You could write the data payload a hex buffer and parse it in javascript directly. In such case, you fully control the characters being injected in the HTML source.

microbit-matt-hillsdon:
I think we should do this if we can. @finneyj what's the background to the comment approach? Is it necessary that we have only a HTML "header" and no trailer?

pelikhan:
What happens when the user insert <, >, or - ? Can you guarantee there is no way to escape the comment sandbox?

pelikhan:
Writing the encoded data as a hex buffer prevents any class of javascript injection that you are exposing yourself with writing user data in a HTML comment.

finneyj:
[...]

PR #111, opened 2021-05-21
MicrobitLog HTML updates

microbit-matt-hillsdon:
Introduces a build step so we can store some reasonable HTML in the
repository and munge it into the C++ array.
    
The online mode uses the dl.css/js files hosted at microbit.org. They're
live, but we might still change the contract between the page and the
hosted files based on feedback.
    
The iframe-based reload will reload the page if it detects that it
changes. The process uses postMessage to notify that a reload is needed
based on a simple hash of the contents.


microbit-matt-hillsdon:
If you test with something like:

```
    for (int i = 0; i < 100; ++i) {
        uBit.log.beginRow();
        uBit.log.logData("index", i);
        uBit.log.logData("a.x", uBit.accelerometer.getX());
        uBit.log.endRow();
        uBit.sleep(1000);
    }
```

then you can see reloads working as you disconnect/reconnect the micro:bit.

microbit-matt-hillsdon:
Misc notes on the JavaScript for me to follow up on:

- I don't trust the comment approach especially as it seems to contain binary data. I found I had to add charset for it to work reliably but still not confident.
- Discussion on #108
- The code to copy assumes that we can create TSV from CSV by replacing commas with tabs. Does the CSV writing allow escaped commas to be written out (e.g. quoted) ? If so it'll break them. CSV formats are a bit of a nightmare when it comes to escaping/quoting and consistency between tools.

finneyj:
I understand concerns about the trailer, but the content is controlled. All data is not uncontrolled binary data. The only things we'll see in there are text or binary bytes with a value of 0xFF (unused data) or 0x00 (invalidated data). if it were uncontrolled, I would completely agree with you.

There is no escaping of characters at all, as I didn't want to use that much flash for the JS code. Again, I was going to filter out commas, and double quotes from keys and values...

resources/logfs/dl.css
```diff
@@ -0,0 +1,42 @@
...
+  border: 2px solid #000;
+  background: #fff;
+}
+tr:hover {
+  background-color: #ddd;
+}
+header,
+footer {
+  /* We can't afford the bytes to have a wrapper element and better CSS in the main doc
+     so use -ve margin to get a full width header. */
```
microbit-matt-hillsdon, line 31: This makes no sense now I've simplified the header styling.

Remove assuming we don't reinstate a black strip header after discussions today.

microbit-matt-hillsdon: We've also got a wrapper element now so can be tidied either way.

resources/logfs/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "codal-microbit-v2-logfs",
+  "version": "0.1.0",
+  "description": "Built tooling for HTML resource",
```
microbit-matt-hillsdon, line 4:
```suggestion
  "description": "Build tooling for HTML resource",
```

microbit-matt-hillsdon:
I went through this with @jaustin earlier and I've implemented all his feedback. @finneyj I think this is ready to review.

The dl.css and dl.js files for the online mode are now on microbit.org.

Tested on Chrome/Safari/FireFox on Mac and Windows 10 Chrome/Edge/IE11 for basic functionality. 

I'll look at a log space indicator as a separate PR.

I'll also look more carefully tomorrow at the C++ changes intended to address the comment escape issue and retest the iframe reload on Windows.

resources/logfs/header.html
```diff
@@ -1,7 +1,148 @@
...
+          // Find the CSV text in the file system
+          let p = d.querySelector("#w");
+          let raw = d.documentElement.outerHTML.split("FS_START")[2];
+          // See MicroBitLog.h/cpp for the format.
+          if (/^UBIT_LOG_FS_V_001/.test(raw)) {
+            let dataStart = parseInt(raw.substr(29, 10), 16) - 2048;
+            let dataSize = 0;
+            while (raw.charCodeAt(dataStart + dataSize) != 0xfffd) {
+              dataSize++;
+            }
```
finneyj, line 80: Does your minifier take out the superfluous braces here?

resources/logfs/header.html
```diff
@@ -1,7 +1,148 @@
...
+            csv = raw.substr(dataStart, dataSize);
+
+            // Hash the content and reload if it changes using an iframe to check for a different hash
+            let hash = 0;
+            for (let i = 0; i < raw.length; ++i) {
+              hash = 31 * hash + raw.charCodeAt(i);
+              hash |= 0;
```
finneyj, line 87: This looks like it does nothing? But is presumably a Javascript-ism?

finneyj:
Looks great @microbit-matt-hillsdon!

finneyj:
Just tested locally, and it works a charm. I really like the auto page refresh feature when the microbit remounts! Very tasty indeed.

I'm happy to merge as soon as you sign off on the minor comments above. Thanks @microbit-matt-hillsdon!

resources/logfs/header.html
microbit-matt-hillsdon: Yep. It ends up as a for loop with no braces.

finneyj: nice!

resources/logfs/header.html
microbit-matt-hillsdon: Hah, yes, that must look weird. It's for the side-effect of converting to a 32 bit integer. JavaScript numbers are 64 bit floats.

finneyj: hehe - thanks. Will try to remember that one!

finneyj:
I think we've got all the input on this now Matt, so let's merge. 

Thanks @microbit-matt-hillsdon!

PR #114, opened 2021-06-05
BLE - make Device Info Service Model Number String sensitive to board ID

martinwork:
MicroBit::init() passes board ID from power.getVersion() to MicroBitBLEManager::init(), which replaces fixed "V2.0" with "V2.0" for known IDs (0x9903/4), and "V2.X" for unknown IDs.


finneyj:
Thanks @martinwork. Looks good.

Issue #110, opened 2021-05-19
MicroBitLog header improvements


jaustin:
We'd like to make some simple changes to the micro:bit log.

Log lives here right now https://github.com/lancaster-university/codal-microbit-v2/tree/master/resources/logfs

## The log has two distinct 'modes'
Offline:
- A simple file to get the data out. Minimum styling (to save flash)

Online:
- Loads some resources from microbit.org to augment the page (css, js)
- Initially this shouldn't add features but should style the page and if possible include more descriptive dialogs

### The changes required are
* Make the online mode exist
* Create an auto-update mechanism
* Style/tweaks to offline mode
* Fix #108 

### Optional
* Improve the 'pipeline' that creates the minified HTML/JS as described in https://github.com/lancaster-university/codal-microbit-v2/tree/master/resources/logfs
* 'space used' indicator that tells the user what percentage of the storage is left (could be on a page level...). This could be in online mode only.
 

## Autoupdate

In both modes we'd  like to use something like the following mechanism to 'auto-update' the data contents when the micro:bit de/re-enumerates
- In a hidden iFrame, periodically reload the MY_DATA.HTM file and use a mechanism (eg post messaging) to determine that the file contents of the two file are (not) the same. If contents do not match, reload ourselves to show the latest content.
- it's likely that comparing the header metadata will do a good job of this, as all writes (currently) result in the length getting incremented in the header block. However @finneyj and I discussed moving to the 'length' field becoming a page pointer and relying on the trailing 0xFFs to determine the end of the data in the relevant block. In this case, the flash page AND the last !0xFF address would be a good proxy for 'is the data different' mechanism. @finneyj could you recommend the smallest/simplest 'hash' for identifying changed data (rare/unlikely collisions are fine...)
 
## Style changes:
* heading should be "micro:bit data log" (IE title case)
* Button titles should be title case, and ellipsis on the end of the two buttons that popup dialogs (or some 'i' icon to show they're information not actions).
* horizontal spacing between the buttons (just a single space kinda works but feels like a hack, but it is only 4 bytes!)
* Vertical space between buttons and table start
* In the table: headers should be center aligned, data right aligned

## For the 'online hydrated' page
* micro:bit logo and simple branding @microbit-matt-hillsdon please do something simple for now and we'll review
* buttons on brand (black, rounded edges)

Do we want to do something more intelligent with the table too? I think it's quite effective as is and there's a benefit to online/offline looking as similar as possible.

[...]

PR #118, opened 2021-06-11
Can't use original text as might still be "Copied"

microbit-matt-hillsdon:
If the 1s timeout hadn't passed then we'll be reinstating the wrong
text for subsequent events, easy to trigger with a double click.

No change to the C++ resource as this feature is in the online HTML
only.

See https://github.com/lancaster-university/codal-microbit-v2/issues/110#issuecomment-859475124

### DIFF
--- a/resources/logfs/dl.js
+++ b/resources/logfs/dl.js
@@ -22,10 +22,9 @@
       copy: function () {
         base.copy();
         let button = window.event.currentTarget;
-        let original = button.innerText;
         button.innerText = "Copied";
         setTimeout(function () {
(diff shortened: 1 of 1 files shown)


microbit-matt-hillsdon:
You can test this with the file built with `npm run test` in this folder. It uses local references to the dl.{js,css} files that are usually on microbit.org. I'll update the microbit.org files once this is merged.

microbit-matt-hillsdon:
CC @jaustin for review

finneyj:
thanks @microbit-matt-hillsdon!

PR #129, opened 2021-06-21
Update datalog microbit.org URLs.

microbit-matt-hillsdon:
The /1/ versions were served with a cache-control header that means the
~3k downloads so far will be cached indefinitely, which limits our
ability to deliver new features. This has since been fixed, but for the
benefit of users who have been testing this, we want to bump the
version.

Content at /dl/1/ and /dl/2/ is the same on microbit.org.


finneyj:
Thanks @microbit-matt-hillsdon LGTM.

PR #139, opened 2021-07-30
Codal.json build flag to support Nordic style BLE UART endpoints

PaulAustin:
A build flag that will swap the endpoint characteristic descriptors.
This enables codal based microbit programs to communicate with apps (iOS/Android/WebBLE) designed to work with NRF5x MBED library conventions.

The option is set with the following flag in the codal.json file. Omitting the flag or setting it to 0 will use conventional microbit settings
{
    "config":{
         "MICROBIT_BLE_NORDIC_STYLE_UART" : 1
   }
}


finneyj:
Thanks @PaulAustin. 

Looks good, and a very useful contribution. Thank you!

I can't quite remember why this is the opposite polarity than the current mbed/nordic apps... I *think* it came about because Nordic flipped the UUIDs soon after we released the v1 micro:bit. Anyhow, a good solution i think!

Issue #153, opened 2021-11-25
`uBit.power.getPowerConsumption()` needs to be updated for the new format


microbit-carlos:
From the latest version of the spec: https://github.com/microbit-foundation/spec-i2c-protocol/blob/main/spec/index.md

![image](https://user-images.githubusercontent.com/29712657/143481338-65b1abe7-3b3b-4895-a0f9-c44e174c7e40.png)

https://github.com/lancaster-university/codal-microbit-v2/blob/9305a13fa4ec3007f70e5024d37f0965ee485444/source/MicroBitPowerManager.cpp#L147-L154

JohnVidler:
This is now fixed as of commit 26e4f565d794b57528b173d0ba5e4d3d9d093692 which provides a new `getPowerData()` call with both returns, and sets `getPowerConsumption()` to be a deprecated proxy for the new method.

PR #154, opened 2021-12-03
Fix UIPM sizes

martinwork:
I think this is strictly correct. If we're not interested in the 2nd word of MICROBIT_UIPM_PROPERTY_POWER_CONSUMPTION, then I think MICROBIT_UIPM_MAX_BUFFER_SIZE should be 7 rather than 8.


microbit-carlos:
Thaks Martin!
This partially addresses https://github.com/lancaster-university/codal-microbit-v2/issues/153.

JohnVidler:
This is addressed in 26e4f565d794b57528b173d0ba5e4d3d9d093692 - so going to close this now

PR #159, opened 2022-01-12
Minor fix to read back the right amount of I2C data from the interface MCU.

microbit-carlos:
A bit inconsequential as the data read back is not being used, but the interface firmware is meant to send back the same number of bytes that was written to it.


JohnVidler:
Agreed on it being tiny, but its important to align with the calling/caller spec. Happy to merge for v0.2.39

PR #166, opened 2022-01-28
Increase DAPLink I2C wait for data to be ready timeout.

microbit-carlos:
The current values for the retries (5 loops => 5ms) is enough for normal operation. However, when MakeCode is WebUSB connected to the micro:bit it must be performing background operations in DAPLink that might delay how long it takes for data to be ready to be sent back to CODAL.

In the worse case scenarios testing with a lot of serial data + requesting ADC readings it has sometimes taken up to 8 retries for the combined interrupt signal to be activated by DAPLink to signal that the data ready. Increasing this value from 5 to 10 shouldn't have any other adverse consequences, other than blocking the fiber in awaitUIPMPacket() a bit longer.

### DIFF
--- a/inc/MicroBitPowerManager.h
+++ b/inc/MicroBitPowerManager.h
@@ -44,7 +44,7 @@ DEALINGS IN THE SOFTWARE.
 // General constants 
 //
 #define MICROBIT_UIPM_MAX_BUFFER_SIZE               8
-#define MICROBIT_UIPM_MAX_RETRIES                   5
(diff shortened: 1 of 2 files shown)


microbit-carlos:
@finneyj @JohnVidler feel free to retarget this PR to the `daplink-interface-versioning` branch if you prefer these DAPLink i2c coms changes to go to main branch all together.

microbit-carlos:
I had to increate this value again, this time to 20, as after further testing the old value wasn't as affective as I thought.

PR #178, opened 2022-03-15
MicroBitUtilityService - to read MicroBitLog MYDATA via BLE

martinwork:
This depends on and includes PR https://github.com/lancaster-university/codal-microbit-v2/pull/177 "Add MicroBitLog::readData".

**TODO**: How should the service and characteristic UUIDs determined?
This currently uses the micro:bit base UUID plus 1 and 2.

MicroBitUtilityService.h/cpp - BLE service providing replies to simple client requests
MicroBitUtilityTypes.h - Details of the service requests that support reading MicroBitLog data
MicroBitCompat.h - Define MICROBIT_ID_UTILITY for service events
MicroBitConfig.h - Macros to control creation in pairing / application mode, default pairing mode only.
MicroBit.cpp - Create service according to config macros. To create in BLEManager::init would need to pass MicroBitLog.
MicroBitBLEChar.cpp - Changed DMESGF to DMESG.

Example hex [log-read-data-ble.hex.zip](https://github.com/lancaster-university/codal-microbit-v2/files/8257001/log-read-data-ble.hex.zip) is built from https://github.com/martinwork/microbit-v2-samples/tree/log-read-data-ble.

### DIFF
--- a/inc/MicroBitConfig.h
+++ b/inc/MicroBitConfig.h
@@ -246,6 +246,18 @@ extern uint32_t __data_end__;
 #define MICROBIT_BLE_DEVICE_INFORMATION_SERVICE 1
 #endif
 
+// Enable/Disable BLE Service in pairing mode: MicroBitUtilityService
+// Set '1' to enable.
+#ifndef MICROBIT_BLE_UTILITY_SERVICE_PAIRING
+#define MICROBIT_BLE_UTILITY_SERVICE_PAIRING 1
+#endif
+
+// Enable/Disable BLE Service in application mode: MicroBitUtilityService
+// Set '1' to enable.
+#ifndef MICROBIT_BLE_UTILITY_SERVICE
+#define MICROBIT_BLE_UTILITY_SERVICE 0
+#endif
+
 // Enable/Disable Nordic Firmware style BLE based UART implimentation.
 // The default codal implimentation reverses the TX/RX ids  
 // Set to '1' to enable
--- a/inc/MicroBitLog.h
+++ b/inc/MicroBitLog.h
@@ -182,6 +182,14 @@ namespace codal
         }
     };
 
+
+    enum class DataFormat
+    {
+        HTMLHeader = 0,   // The HTML header without the data
+        HTML = 1,         // The entire HTML file
+        CSV = 2           // CSV data
+    };
+
     /**
      * Class definition for MicroBitLog. A simple text only, append only, single file log file system.
      * Also contains a key/value pair abstraction to enable dynamic creation of CSV based logfiles.
@@ -326,6 +334,30 @@ namespace codal
          * @param s the string to inject.
          */
         int logString(ManagedString s);
+        
+        /**
+         * Get the length of the recorded data
+         * @param format the data format
+         *          DataFormat::HTMLHeader = 0,   - The HTML header without the data
+         *          DataFormat::HTML = 1,               - The entire HTML file
+         *          DataFormat::CSV = 2                   - CSV data
+         * @return the length of the recorded data
+         */
+        uint32_t getDataLength(DataFormat format);
+        
+        /**
+         * Read the recorded data
+         * @param data pointer to memory to store the data
+         * @param index  the index into the data
+         * @param len length of the data to fetch
+         * @param format the data format
+         *          DataFormat::HTMLHeader = 0,   - The HTML header without data
+         *          DataFormat::HTML = 1,               - The entire HTML file with data
+         *          DataFormat::CSV = 2                   - CSV data
+         * @param length expected complete length returned from getDataLength
+         * @return DEVICE_OK on success; DEVICE_INVALID_PARAMETER if data is not available for the request
+         */
+        int readData(void *data, uint32_t index, uint32_t len, DataFormat format, uint32_t length);
 
     private:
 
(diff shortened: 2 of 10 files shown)


JohnVidler:
So I've done some digging, and for UUIDs it seems that we can either:
1. Register with the Bluetooth folks for a specific set of numbers, at extreme cost
2. Use pre-existing ones from the giant list here: https://btprodspecificationrefs.blob.core.windows.net/assigned-values/16-bit%20UUID%20Numbers%20Document.pdf
3. Just make some up (what you've got here 👍 ) and just have magic micro:bit numbers.

3 is nice and simple, but is frowned upon my the Bluetooth folks, and there are _a lot_ of UUIDs in the 16-bit table, so there may well be an appropriate one for us to use, but scanning them right now I haven't spotted one yet (but the list is massive).

If someone fancies like scouring the UUID list I'd appreciate it, as I have to look at other issues for v0.2.40 and this is a big time sink :(

Otherwise if we have time before zero-hour, I'll see what I can find and see if we can get this and #177 merged in for v0.2.40.

martinwork:
> Just make some up (what you've got here 👍 ) and just have magic micro:bit numbers.

I've always assumed micro:bit has a registered base UUIID, so I didn't expect we would need to register a new one:
https://github.com/lancaster-university/codal-microbit-v2/blob/master/source/bluetooth/MicroBitBLEService.cpp#L44

Many of the existing micro:bit specific UUIDs combine this base ID with various 16 bit IDs
https://github.com/lancaster-university/codal-microbit-v2/blob/master/source/bluetooth/MicroBitEventService.cpp#L40
https://github.com/lancaster-university/codal-microbit-v2/blob/master/source/bluetooth/MicroBitIOPinService.cpp#L37

I don't know how those 16bit IDs were picked. They look random, maybe to minimise clashes with users experimenting with their own services.

JohnVidler:
Bumped to *41, as its too close to release to keep now with undefined IDs.

bsiever:
By the way, I think just getting the data from `readData()` leaves a few things out.  I included: 

1. Getting the current micro:bit time (to correlate log data with wall clock time --- there are some limitations doing this, but nothing like it is easily possible with the current data logger.  Well, unless you use an RTC extension. 
2. A modest notion of "security".  I use a simple passphrase to authenticate.  It's a non-encrypted link and could be sniffed out, but this approach seems like a good balance between the some optional security, ease of use, and the minor risks posed.
3.  The ability to remotely erase the log.
4. An estimate of the total log space currently used. 

Things that would be really nice:  a _rolling_ log (when full, just overwrite the earliest entries).  This requires some additional  thought to the logging approach.  I think my initial ideas all require a little more wear on flash in ways.  For example, use sentinel values that represent log start, stop, or the combined start stop if the log has rolled over.  Each write would have to do the data and write/update the sentinel.   On the other hand, the current approach will always start at the beginning of the designated after any `clear()`.  The rolling approach could start at wherever the start sentinel is, so it may wear the flash more uniformly.  A `clear()` would really just be an update of the end sentinel value and not reset writes to the beginning.

jaustin:
@bsiever I like the ability to get the micro:bit time, but I wonder if MicroBitUtilityService and some equivalent Web USB service/protocol should also allow us to *set* wall clock time on the micro:bit. Certainly when playing with some RTC-based addon boards recently that felt useful.

For security, why not use the encryption options provided by Bluetooth, like turning on passphrase pairing? This seems better than doing roll-our-own which is always going to be fairly easily worked around?

Erasing the log without user action seems a bit suspect, but I guess this would allow us to setup a case where the micro:bit is out of reach or deeply embedded in something?

Do you need to be able to estimate log space available without reading the whole file?

bsiever:
@jaustin 

## Setting time 

Current Time Service (https://www.bluetooth.com/specifications/specs/current-time-service-1-1/) may be an ok fit, but probably more overhead than needed.  I'll leave that to you and the rest of the micro:bitters (hmm...that can't be the right term).

## Security

I'm definitely open to thinking more about security....but:
* Pairing with "Just works" doesn't really secure the data. Anyone can pair. 
* Pairing with a passphrase requires considering all the devices one wants to pair with in advance of deployment. 
  * Setup takes time 
  * Reprogramming / redeploying or upgrades can sometimes reset the keys on just one end.  This is frustrating to users who went through pairing and don't know why it's not connecting.
  * Teachers may use this very intermittently.  I don't want someone who does this once a year to have to think about "Did I pair this micro:bit?  To this computer? It is still good?".  
  * I want it to be easy to have multiple sensors collecting data.  Deploy the same code to 5 sensors.  Dump them somewhere. Three days later walk by, open an app, and get data from all 5 with the same passphrase.  Minimal effort.  An example suggested by a colleague:  Collect temperature at several places on a playground.  See the impact of shadow, day/night, shelter from the wind, etc. 

For the bulk of use cases I have in mind, no security or a simple password seems easier and doesn't carry much risk.   

## Erasing
* Erasing does require a passphrase if one was selected. 
* I wouldn't bother if there was a rolling log.  The main concern is long term deployments that will fill the log. 

## Log Space 

It'd be nice if CODAL did support an estimate of remaining log, but here's how I'm currently doing it (not well tested): https://github.com/bsiever/microbit-pxt-blelog/blob/754578c06e0f3bbcfbd7f702454a5865fe0e8268/BLELogService.cpp#L250 .

martinwork:
Created https://github.com/lancaster-university/codal-microbit-v2/pull/287 to replace this one.

microbit-carlos:
Closing this PR in favour of continuing the work in:
- https://github.com/lancaster-university/codal-microbit-v2/pull/287

PR #185, opened 2022-05-04
Add polyphonic real-time synthesizer.

preglow:
Disabled by default.


JohnVidler:
As these are isolated classes, I'm happy to just merge this.

PR #190, opened 2022-06-09
Added tentative support for the BUSY flag on the KL27

JohnVidler:
I'm currently unable to test this, but this incorporates the same logic as the USB Flash Manager to detect and handle KL27 BUSY flag operations when waiting for data back over i2c.

@microbit-carlos Can you confirm if this is working on the latest DAPLink please.


JohnVidler:
Tested to my own satisfaction on a v2.00 board with a beta DAPLink version to support the flag - all seems well, so squashing and merging

PR #192, opened 2022-06-09
Removed old accelerometer and compass driver

JohnVidler:
Removed the old accelerometer and compass drivers, minimised auto-detection and simplified the interface (now just a thin proxy).

Ultimately we want to make this completely static, but for the next release this is an acceptable code-path squash :)

### DIFF
--- a/source/MicroBitAccelerometer.cpp
+++ b/source/MicroBitAccelerometer.cpp
@@ -24,19 +24,17 @@ DEALINGS IN THE SOFTWARE.
 
 #include "MicroBitAccelerometer.h"
 #include "MicroBitIO.h"
-#include "ErrorNo.h"
 #include "MicroBitEvent.h"
 #include "MicroBitCompat.h"
 #include "MicroBitFiber.h"
 #include "MicroBitDevice.h"
 #include "MicroBitI2C.h"
 #include "MicroBitCompass.h"
-#include "FXOS8700.h"
 #include "LSM303Accelerometer.h"
 #include "LSM303Magnetometer.h"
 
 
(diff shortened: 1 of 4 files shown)


source/MicroBitAccelerometer.cpp
```diff
@@ -71,34 +68,21 @@ Accelerometer& MicroBitAccelerometer::autoDetect(MicroBitI2C &i2c)
...
-        // All known accelerometer/magnetometer peripherals have the same alignment
-        if (FXOS8700::isDetected(i2c, 0x3E))
-        {
-            FXOS8700 *fxos = new FXOS8700(i2c, irq1, coordinateSpaceFXOS8700, 0x3E);
-            MicroBitAccelerometer::detectedAccelerometer = fxos;
-            MicroBitCompass::detectedCompass = fxos;
-        }
+        // Now, probe for the LSM303, and if it doesn't reply, panic
+        if (!LSM303Accelerometer::isDetected(i2c, 0x32))
+            microbit_panic( 50 ); // FixMe: This should be a constant from the PanicCode enum, but this is out of sync
```
martinwork, line 76: @JohnVidler I think DAL delays the panic until the accelerometer is used (https://github.com/lancaster-university/microbit-dal/commit/96126fd23f3147257c62cd1aa98a26588c869d63), so that an unresponsive accelerometer might not stop all programs running. I'm not sure if this idea was implemented in CODAL. It looks like CODAL simply didn't panic.

JohnVidler: Good catch, this is now addressed in subsequent patches.

Issue #143, opened 2021-08-16
datalog: When log is empty visual preview throws unknown error


microbit-carlos:
We could check this use case and provide a better error:

![image](https://user-images.githubusercontent.com/29712657/129597010-bb122b51-477f-432d-8ade-a7c7b21ed256.png)

microbit-carlos:
The online JS can first check if there is any data, and if it's empty we can grey out the visualise button.

microbit-carlos:
Fixed in https://github.com/lancaster-university/codal-microbit-v2/pull/198 🎉

PR #198, opened 2022-06-14
datalog htm: Fix unintuitive error msg when trying to plot with empty data.

microbit-carlos:
Fixes issue https://github.com/lancaster-university/codal-microbit-v2/issues/143.

A simple one in the online JS, basically if the log is empty and the user clicks on "Visual Preview" it showed an `Unknown error` or `The visual preview requires two or more columns. Timestamps must be enabled.`

So, with this PR it will show the expected message `No data to graph.`.

### DIFF
--- a/resources/logfs/dl.js
+++ b/resources/logfs/dl.js
@@ -112,7 +112,7 @@ function UserGraphError(message) {
 
       graph: function () {
         function readData(table) {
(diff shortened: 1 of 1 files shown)


JohnVidler:
Checked and happy to merge now, sorry this one got missed for 0.2.40 :/

Issue #213, opened 2022-08-10
Panic 50/51 for the motion sensors should be thrown on first use


microbit-carlos:
At the moment I believe if there is an issue communicating with the motion sensor then we immediately panic with error code 050/051.

To match the V1 functionality, we should delay the panic to the first usage of the motion sensor.

This is not directly related, but in the same area as:
- https://github.com/lancaster-university/codal-microbit-v2/issues/191

martinwork:
See https://github.com/lancaster-university/codal-microbit-v2/pull/192#pullrequestreview-1003021469

JohnVidler:
This is correcly handled in `master` and will be the case for tags from `0.2.41` onwards.

The implementation allows for a weird kind of partial failure in the case of the Micro:bit v2 where our compass and accelerometer are the same physical device, and includes support for the compass (for example) failing to come up, but the accelerometer working just fine, but I've left it this way in case we have any extensions to either class in the future, or on other boards, so they can independently fail.

See also:
- https://github.com/lancaster-university/codal-microbit-v2/issues/191#issuecomment-1213239668

microbit-carlos:
As part of this ticket we should also check if there is a way to simplify a bit, or reduce how many places throw the panic, as discussed in:
- https://github.com/lancaster-university/codal-microbit-v2/pull/216#issuecomment-1216625005

Issue #191, opened 2022-06-09
Panic Codes don't match the micro:bit V1


microbit-carlos:
V1 defined in the DAL repo: https://github.com/lancaster-university/microbit-dal/blob/602153e9199c28c08fd2561ce7b43bf64e9e7394/inc/core/ErrorNo.h#L72-L89

```cpp
enum PanicCode{
    // PANIC Codes. These are not return codes, but are terminal conditions.
    // These induce a panic operation, where all code stops executing, and a panic state is
    // entered where the panic code is diplayed.

    // Out out memory error. Heap storage was requested, but is not available.
    MICROBIT_OOM = 20,

    // Corruption detected in the micro:bit heap space
    MICROBIT_HEAP_ERROR = 30,

    // Dereference of a NULL pointer through the ManagedType class,
    MICROBIT_NULL_DEREFERENCE = 40,

    // A requested hardware peripheral could not be found,
    MICROBIT_HARDWARE_UNAVAILABLE_ACC = 50,
    MICROBIT_HARDWARE_UNAVAILABLE_MAG = 51,
};
```

V2 might be using the values from the codal-core repo:
https://github.com/lancaster-university/codal-core/blob/befa2c51b3595ca0d44e1bc5daad0531974b77e7/inc/core/ErrorNo.h#L80-L105

```cpp
enum PanicCode{
    // PANIC Codes. These are not return codes, but are terminal conditions.
    // These induce a panic operation, where all code stops executing, and a panic state is
    // entered where the panic code is diplayed.

    // Out out memory error. Heap storage was requested, but is not available.
    DEVICE_OOM = 20,

    // Corruption detected in the codal device heap space
    DEVICE_HEAP_ERROR = 30,

    // Dereference of a NULL pointer through the ManagedType class,
    DEVICE_NULL_DEREFERENCE = 40,

    // Non-recoverable error in USB driver
    DEVICE_USB_ERROR = 50,

    // Non-recoverable error in the JACDAC stack
    DEVICE_JACDAC_ERROR = 60,

    // CPU SDK faults
    DEVICE_CPU_SDK = 70,

    // hardware incorrect configuration
    DEVICE_HARDWARE_CONFIGURATION_ERROR = 90,
};
```

JohnVidler:
@finneyj Do you have some comment on this?

[...]

PR #216, opened 2022-08-12
Added soft-failure modes for the accelerometer and compass to prevent panic-on-boot issues

JohnVidler:
Added a new error class which uses PanicCode as a base for Micro:bit v2 specific errors, and implemented these in the Accel and Compass classes.

This addresses the case where a damaged accel/compass module would cause a board to panic on boot, effectively bricking the device even if the user code never touches the peripheral.

See also:
- https://github.com/lancaster-university/codal-core/pull/151
- https://github.com/lancaster-university/codal-microbit-v2/issues/213
- https://github.com/lancaster-university/codal-microbit-v2/issues/191

### DIFF
--- a/inc/MicroBitError.h
+++ b/inc/MicroBitError.h
@@ -1,0 +1,38 @@
+/*
+The MIT License (MIT)
+Copyright (c) 2022 Lancaster University.
+Permission is hereby granted, free of charge, to any person obtaining a
+copy of this software and associated documentation files (the "Software"),
+to deal in the Software without restriction, including without limitation
+the rights to use, copy, modify, merge, publish, distribute, sublicense,
+and/or sell copies of the Software, and to permit persons to whom the
+Software is furnished to do so, subject to the following conditions:
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+*/
+
+#ifndef MicroBit_Error_H
+#define MicroBit_Error_H
+
+#include "ErrorNo.h"
+
+// Defined Micro:bit v2.xx specific error codes
+// For on-board peripherals these should be from PanicCode::DEVICE_PERIPHERAL_ERROR offset
+// to ensure we're in the correct range.
+// Note: the maximum offset for this range is +9, to avoid hitting the next block!
+enum MicroBitPanic {
+    // Accelerometer unable proceed (hardware i/o failure)
+    ACCELEROMETER_ERROR = PanicCode::DEVICE_PERIPHERAL_ERROR + 1, // Aka. Panic 51
+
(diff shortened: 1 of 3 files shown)


inc/MicroBitError.h
```diff
@@ -0,0 +1,38 @@
...
+// For on-board peripherals these should be from PanicCode::DEVICE_PERIPHERAL_ERROR offset
+// to ensure we're in the correct range.
+// Note: the maximum offset for this range is +9, to avoid hitting the next block!
+enum MicroBitPanic {
+    // Accelerometer unable proceed (hardware i/o failure)
+    ACCELEROMETER_ERROR = PanicCode::DEVICE_PERIPHERAL_ERROR + 1, // Aka. Panic 51
+
+    // Compass unable to proceed (hardware i/o failure)
+    COMPASS_ERROR = PanicCode::DEVICE_PERIPHERAL_ERROR + 2        // Aka. Panic 52
+};
```
microbit-carlos, line 36: To match V1, we need these to be values 50 and 51, not 51 and 52: 
https://support.microbit.org/support/solutions/articles/19000016969-micro-bit-error-codes

microbit-carlos:
Also, the way this was done in DAL was to only have a single place (`requestUpdate()`) that calls the panic function. Is there a single acc/mag  method that will ultimately be called by all other acc/mag methods? That way we'd avoid having an `if` check on each method.

martinwork:
@JohnVidler It took me ages to realise that uBit.accelerometer is a codal::Accelerometer reference initialised to the return value of MicroBitAccelerometer::autoDetect, which is now a codal::LSM303Accelerometer.

LSM303Accelerometer doesn't override getX(), so uBit.accelerometer.getX() calls Accelerometer::getX() which calls the LSM303Accelerometer::requestUpdate() overide.

Like the DAL, when autoDetect fails, it could default to returning a codal::Accelerometer instead of a codal::LSM303Accelerometer, so uBit.accelerometer is still valid. CODAL used to do this I think. Then codal::Accelerometer::requestUpdate() could do the panic.

From uBit, none of the functions like MicroBitAccelerometer::getX() are called.

JohnVidler:
> @JohnVidler have you tested this PR before merging? If I change `LSM303Accelerometer::isDetected()` to always return false and then try to use the accelerometer, there is no panic thrown and it looks like the code freezes somewhere.

Limited testing proved to work on my kit, but I was pressed for time before being on leave. I'll revisit this.

> Like the DAL, when autoDetect fails, it could default to returning a codal::Accelerometer instead of a codal::LSM303Accelerometer, so uBit.accelerometer is still valid. CODAL used to do this I think. Then codal::Accelerometer::requestUpdate() could do the panic.

The only issue with just using `requestUpdate` for the check(s) is that the object also implements `idleCallback` which calls `requestUpdate` regularly, and there's no immediate way to check if the call is from a request for data, OR from an idle update (without adding state).

I'll re-examine this lot shortly and sort a solution; but I had assumed that everything was accessing data through one of the method calls, so we could rely on those to determine if a user-driven access was occuring, but if the object data is just being exposed directly after a `requestUpdate` call, then we can't rely on that.

One option might be to have this soft fail perhaps? Record all zero for the data if the chip cannot be reached... although this has its own problems (clarity for the user for failing hardware, and ease of support)

martinwork:
> The only issue with just using requestUpdate for the check(s) is that the object also implements idleCallback which calls requestUpdate regularly

idleCallback is not enabled until requestUpdate enables it. All external calls are in codal::Accelerometer, and call requestUpdate.
https://github.com/lancaster-university/codal-core/blob/master/source/drivers/LSM303Accelerometer.cpp#L82
https://github.com/lancaster-university/codal-core/blob/master/source/drivers/LSM303Accelerometer.cpp#L157
https://github.com/lancaster-university/codal-core/blob/master/source/driver-models/Accelerometer.cpp#L420

If failure to initialise the LSM303 resulted in uBit.accelerometer being a reference to a codal::Accelerometer (as it used to), then idleCallback would never be enabled, and codal::Accelerometer::requestUpdate (always overridden by a real accelerometer subclass) can panic
https://github.com/lancaster-university/codal-core/blob/master/source/driver-models/Accelerometer.cpp#L389
like the DAL
https://github.com/lancaster-university/microbit-dal/blob/master/source/drivers/MicroBitAccelerometer.cpp#L442

It's super confusing that MicroBitAccelerometer is the base class in the DAL, but not in a micro:bit CODAL build.

Issue #204, opened 2022-07-01
Datalog: Simple hello world CODAL programme doesn't erase data in datalog


microbit-carlos:
- First flash this MicroPython datalogging programme, press the buttons a couple of times and confirm there is data in the log
    - [uPy-datalogging.hex.zip](https://github.com/lancaster-university/codal-microbit-v2/files/9030916/uPy-datalogging.hex.zip)
- Then flash this CODAL programme using v0.2.40
    - [CODAL-helloworld-only.hex.zip](https://github.com/lancaster-university/codal-microbit-v2/files/9030920/CODAL-helloworld-only.hex.zip)
- Unplug and replug the micro:bit drive a couple of times to make sure CODAL had a chance to erase the datalog
- The MY_DATA.HTML still contains the old data

```cpp
#include "MicroBit.h"

MicroBit uBit;

int  main() {
    uBit.init();

    while (true) {
        uBit.display.scroll("HELLO WORLD!");
    }
}
```

Related to:
- https://github.com/lancaster-university/codal-microbit-v2/issues/182

JohnVidler:
@microbit-carlos It might just be (from looking through the erase logic) that we just need `CONFIG_MICROBIT_ERASE_USER_DATA_ON_REFLASH` by default in `codal.json` for new projects... I'll test this and if its the case we can perhaps just bump this flag into the default `codal.json` config.

martinwork:
@JohnVidler @microbit-carlos
CONFIG_MICROBIT_ERASE_USER_DATA_ON_REFLASH is defaulted to 1 here
https://github.com/lancaster-university/codal-microbit-v2/blob/13a651754e668464855fc40a81dd7dcb796b58c9/model/MicroBit.h#L107

The problem is that, when MicroBitLog::invalidate() is called, MicroBitLog::_IsPresent() is returning false here
https://github.com/lancaster-university/codal-microbit-v2/blob/897e18b11fdfbd00123db0b435b949e7ad95e8e5/source/MicroBitLog.cpp#L973
because the python program is storing
MICROBIT_LOG_VERSION = UBIT_LOG_FS_V_001
but the newly compiled CODAL program is looking for
MICROBIT_LOG_VERSION = UBIT_LOG_FS_V_002.

JohnVidler:
Not 100% sure how to proceed on this one; do we want to always reset the log _irrespective of version_, and if so, then the fix is to check via `memcmp(..., 14)` (skipping the actual version number).

[...]

Issue #182, opened 2022-04-28
datalog: Running a programme without datalogging does not remove MY_DATA.HTM file


microbit-carlos:
To replicate:

- Drag&drop an `erase.act` file into the `MICROBIT` drive while pressing down the reset button
    - This will fully erase the DAPLink storage and the target programme
- Build and flash simple programme using datalogging, for example: [MICROBIT_datalogging.hex.zip](https://github.com/lancaster-university/codal-microbit-v2/files/8583513/MICROBIT_datalogging.hex.zip)

  ```cpp
  #include "MicroBit.h"

  MicroBit uBit;

  int main() {
      uBit.init();
      uBit.log.setVisibility(true);
      while(1)
          uBit.display.scroll("data logging");
  }
  ```
- The `MY_DATA.HTM` file should now be in the `MICROBIT` drive
- Build and flash a simple programme without datalogging, for example: [MICROBIT_hello-world.hex.zip](https://github.com/lancaster-university/codal-microbit-v2/files/8583537/MICROBIT_hello-world.hex.zip)

  ```cpp
  #include "MicroBit.h"

  MicroBit uBit;

  int  main() {
      uBit.init();
      while(1)
          uBit.display.scroll("HELLO WORLD!");
  }
  ```
- Power cycle the board a couple of times
- The `MY_DATA.HTM` file is still in the `MICROBIT` drive

microbit-carlos:
Ideal resolution: When data logging is not used the file does not appear on the MICROBIT drive from the moment is mounted.

We want to make sure drive remounts are reduced as much as possible, so we'd like to only remount when transitioning between "programmes that have datalogging" and "programmes that do not have datalogging".
So if a user is constantly loading a similar programme, with or without datalogging, the drive should not remount, as it is not "changing datalogging state".

microbit-carlos:
In a conversation with @finneyj and @JohnVidler we believe there is some code in CODAL that is meant to be removing the MY_DATA.HTM file from the MICROBIT drive, but it might not be working as expected.

PR #217, opened 2022-08-12
Modification to MICROBIT_LOG_VERSION checks for _isPresent() calls

JohnVidler:
MicroBitLog now only checks the version string preamble, to avoid bugs where old logs are not detected during an erase cycle.

This effectively just checks the first 14 bytes of `MICROBIT_LOG_VERSION` aka `UBIT_LOG_FS_V_` such that a log written with `...001` is still detected as present by code running with `...002`, and will still be erased during a reflash, for example.


JohnVidler:
In theory this would close #204 and #182, but as this needs testing I've elected to merge it in for Beta release, then we can check before going live, _then_ close the other tickets.

Issue #203, opened 2022-06-27
MakeCode power extension programme does not go to sleep 


microbit-carlos:
## Description

This programme does seem to go to sleep.
Although it's possible it goes to sleep but something wakes it up immediately.

Changing the `[request low power]` for a `[request low power for [10000] ms` block and it seems to work.

## Test Programme

<img width="645" alt="image" src="https://user-images.githubusercontent.com/29712657/176009864-1769c6a0-8cdf-4c21-862d-b13c8f0bf73d.png">

```ts
input.onButtonPressed(Button.A, function () {
    basic.showString("A")
    power.lowPowerRequest()
})
basic.forever(function () {
    basic.showNumber(0)
    basic.showIcon(IconNames.Heart)
})
```

## Hardware & Software

Tested on a V2.2 using the micro:bit power management extension and MakeCode live.

MakeCode info:
- makecode.microbit.org version:  [5.0.6](https://github.com/Microsoft/pxt-microbit/releases/tag/v5.0.6)
- Microsoft MakeCode version:  [8.0.3](https://github.com/Microsoft/pxt/releases/tag/v8.0.3)
- microbit runtime version:  [v2.2.0-rc6](https://github.com/lancaster-university/microbit/releases/tag/v2.2.0-rc6)
- codal-microbit-v2 runtime version:  [v0.2.40](https://github.com/lancaster-university/codal-microbit-v2/releases/tag/v0.2.40)

Extension:
- https://github.com/microbit-foundation/pxt-microbit-v2-power
- v0.2.3

Share link and hex:
- https://makecode.microbit.org/_eEL9T4CEear
- [microbit-Pwr-test-sleep.hex.zip](https://github.com/lancaster-university/codal-microbit-v2/files/8994684/microbit-Pwr-test-sleep.hex.zip)
m

JohnVidler:
Placing this in the v0.2.41 release for now, mainly to facilitate tracking it

martinwork:
The power manager blocks low power mode if there are no wake up sources or if a timed wake up is due to trigger soon. A button handler does not automatically create a wake up source. See https://makecode.microbit.org/_Xwxhku8zqVUm. The forever loop has a pause because of https://github.com/lancaster-university/codal-microbit-v2/issues/201#issuecomment-1214165440
![image](https://user-images.githubusercontent.com/989126/184498268-f73b3a58-a2f1-4046-9bec-c9c03f763e16.png)

[...]

PR #229, opened 2022-09-20
Bug/deepsleep anyway

JohnVidler:
Addresses the following combined:

- https://github.com/lancaster-university/codal-microbit-v2/issues/203
- https://github.com/microbit-foundation/micropython-microbit-v2/issues/127

I've merged these two together as they both deal with sleep issues, and were developed and tested together.

Also includes the patch from the `hotfix` branch which avoided a hung board if anything used the sound emoji synth from a higher-level language.

All grouped together to get us back on track for continued development from `master`

### DIFF
--- a/model/MicroBit.cpp
+++ b/model/MicroBit.cpp
@@ -119,6 +119,9 @@ MicroBit::MicroBit() :
     // serial.set_flow_control(mbed::Serial::Disabled);
     //serial.baud(115200);
 
+    // Enable the serial port as a deep sleep wake event. (Useful for MicroPython REPL, for example)
+    serial.status |= CODAL_SERIAL_STATUS_DEEPSLEEP;
+
     // Add pullup resisitor to IRQ line (it's floating ACTIVE LO)
     io.irq1.getDigitalValue();
     io.irq1.setPull(PullMode::Up);
--- a/source/MicroBitPowerManager.cpp
+++ b/source/MicroBitPowerManager.cpp
@@ -762,25 +762,6 @@ int MicroBitPowerManager::canDeepSleep( bool wakeOnTime, CODAL_TIMESTAMP wakeUpT
             return DEVICE_INVALID_STATE;
         }
     }
-    /*else
-    {
-        if ( wakeUpSources)
-        {
-            deepSleepCallbackData data;
-            CodalComponent::deepSleepAll( deepSleepCallbackCountWakeUps, &data);
-            if ( data.count == 0)
-            {
-                return DEVICE_INVALID_STATE;
-            }
-        }
-        else
-        {
-            if ( wakeUpPin == NULL)
-            {
-                return DEVICE_INVALID_STATE;
-            }
-        }
-    }*/
 
     return DEVICE_OK;
 }
--- a/source/MicroBitPowerManager.cpp
+++ b/source/MicroBitPowerManager.cpp
@@ -475,10 +475,10 @@ void MicroBitPowerManager::deepSleep()
         return;
     }
 
-    CODAL_TIMESTAMP eventTime = 0;
-    bool wakeOnTime = system_timer_deepsleep_wakeup_time( eventTime);
-    int can = canDeepSleep( wakeOnTime, eventTime, true /*wakeUpSources*/, NULL /*wakeUpPin*/);
-    if ( can == DEVICE_OK)
+    //CODAL_TIMESTAMP eventTime = 0;
+    //bool wakeOnTime = system_timer_deepsleep_wakeup_time( eventTime);
+    //int can = canDeepSleep( false, eventTime, true /*wakeUpSources*/, NULL /*wakeUpPin*/);
+    //if ( can == DEVICE_OK)
         deepSleepWait();
 }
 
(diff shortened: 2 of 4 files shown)


source/MicroBitPowerManager.cpp
```diff
@@ -475,10 +475,10 @@ void MicroBitPowerManager::deepSleep()
...
     }
 
-    CODAL_TIMESTAMP eventTime = 0;
-    bool wakeOnTime = system_timer_deepsleep_wakeup_time( eventTime);
-    int can = canDeepSleep( wakeOnTime, eventTime, true /*wakeUpSources*/, NULL /*wakeUpPin*/);
-    if ( can == DEVICE_OK)
+    //CODAL_TIMESTAMP eventTime = 0;
+    //bool wakeOnTime = system_timer_deepsleep_wakeup_time( eventTime);
+    //int can = canDeepSleep( false, eventTime, true /*wakeUpSources*/, NULL /*wakeUpPin*/);
+    //if ( can == DEVICE_OK)
```
finneyj, line 481: Can we just delete these lines rather than leave dead code?

source/MicroBitPowerManager.cpp
```diff
@@ -762,25 +762,6 @@ int MicroBitPowerManager::canDeepSleep( bool wakeOnTime, CODAL_TIMESTAMP wakeUpT
...
-            }
-        }
-        else
-        {
-            if ( wakeUpPin == NULL)
-            {
-                return DEVICE_INVALID_STATE;
-            }
-        }
-    }
```
finneyj, line 783: Looks fine to me - I think the behaviour of deepsleeping when asked, even if there are no wake up sources, is a good behaviour.

source/SoundEmojiSynthesizer.cpp
```diff
@@ -207,10 +207,7 @@ ManagedBuffer SoundEmojiSynthesizer::pull()
                 done = true;
                 if (renderComplete || status & EMOJI_SYNTHESIZER_STATUS_STOPPING)
                 {
-                    // Wait for a hanful of milliseconds while we finish playing the last buffer we sent out ...
-                    //fiber_sleep( buffer.length() / sampleRate ); // Removed due to causing issues for Python
-
-                    // ... THEN flip out status and fire the event
+                    // Flip out status and fire the event
```
finneyj, line 210: I'm not sure I understand that comment?

finneyj (commented):
thanks @JohnVidler! Looks good to me, pending the minor changes below.

microbit-carlos:
I think there is also an async version of deep sleep, should that also be updated?

source/MicroBitPowerManager.cpp
JohnVidler: Sure, I mostly just left those in as a prompt for me for the old behaviour

model/MicroBit.cpp
```diff
@@ -119,6 +119,9 @@ MicroBit::MicroBit() :
     // serial.set_flow_control(mbed::Serial::Disabled);
     //serial.baud(115200);
 
+    // Enable the serial port as a deep sleep wake event. (Useful for MicroPython REPL, for example)
+    serial.status |= CODAL_SERIAL_STATUS_DEEPSLEEP;
```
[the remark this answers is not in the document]
JohnVidler: Sure we can give it a build var, but I wondered if this was a good one to have an 'opinionated default' on, as:

1. We don't have anything really using the functionality yet ...
3. We're in control of the default behaviour in all three languages, and ...
3. It provides a safety net for locking your board up that requires no updates to any language/editor

Happy to go back to the original behaviour and add a config var if people have strong feelings on this though.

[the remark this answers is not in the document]
finneyj: Well, let's whack in a CONFIG option anyway, then we can discuss / easily change the default value. :)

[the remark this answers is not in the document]
finneyj: I think OFF is the correct default personally, as i's not clear (for the general case) that it should be any different to the other wake up sources. But override that for micropython, where there's a clear use case...

JohnVidler:
> I think there is also an async version of deep sleep, should that also be updated?

Yup, but as we're modifying `canDeepSleep` here and its used by both calls, both are fixed the same way.

source/MicroBitPowerManager.cpp
JohnVidler: Removed: https://github.com/lancaster-university/codal-microbit-v2/pull/229/commits/610bafb4c4c049338babbce6fde4fa4bdcc3d972

source/SoundEmojiSynthesizer.cpp
JohnVidler: Fixed the typo, for clarity: https://github.com/lancaster-university/codal-microbit-v2/pull/229/commits/b5236e0ddc42dac33c9a3779dc6c3fd46812a505

model/MicroBit.cpp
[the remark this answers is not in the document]
JohnVidler: OFF is the default config now (both for individual UART devices, and the whole board) but can be configured via `MICROBIT_USB_SERIAL_WAKE` in `codal.json` to support Python (and other HLLs) who might want thsi feature.

Issue #150, opened 2021-11-05
if a `uBit.sleep()` is used followed by `target_wait_us()` the delays do not match


microbit-carlos:
I expect the 4ms resolution of `uBit.sleep()` as that is the ticker configuration, however it'd expect `target_wait_us()` to not have this limitation as it spin waits using the timer peripheral.

To test via pin toggle with a logic analyser we can first confirm that `target_wait_us()` works fine on its own:

```cpp
while (true) {
    uBit.io.P15.setDigitalValue(1);
    target_wait_us(4000);
    uBit.io.P15.setDigitalValue(0);
    target_wait_us(5000);
    uBit.io.P15.setDigitalValue(1);
    target_wait_us(6000);
    uBit.io.P15.setDigitalValue(0);
    target_wait_us(7000);
    uBit.io.P15.setDigitalValue(1);
    target_wait_us(8000);
    uBit.io.P15.setDigitalValue(0);
    target_wait_us(9000);
}
```

Ignoring the inaccuracy of `target_wait_us` (4ms delay is measured as 3.85ms), we can clearly see the delay increasing proportionaly:
<img width="1546" alt="Screenshot 2021-11-05 at 15 23 07" src="https://user-images.githubusercontent.com/29712657/140535357-c1fed152-2462-48cd-8f91-47f8605e690e.png">

Adding a `uBit.sleep(4)`, as that is minimum delay due to the 4ms ticket, with each `target_wait_us()` and the results change (measurement results added in the comments of this snippet):
```cpp
while (true) {
    uBit.io.P15.setDigitalValue(1);
    // This delay is 7.85ms -> looks "fine", similar inaccuracy as the previous test
    uBit.sleep(4);
    target_wait_us(4000);
    uBit.io.P15.setDigitalValue(0);
    // This delay is 8.95ms -> this looks good too
    uBit.sleep(4);
    target_wait_us(5000);
    uBit.io.P15.setDigitalValue(1);
    // This delay is ~13ms ❌
    uBit.sleep(4);
    target_wait_us(6000);
    uBit.io.P15.setDigitalValue(0);
    // This delay is ~13ms ❌
    uBit.sleep(4);
    target_wait_us(7000);
    uBit.io.P15.setDigitalValue(1);
    // This delay is  ~13ms ❌
    uBit.sleep(4);
    target_wait_us(8000);
    uBit.io.P15.setDigitalValue(0);
    // This delay is  ~13ms -> I guess this is finally correct
[...]

PR #248, opened 2023-01-18
docstrings: Add granularity info to MicroBit::sleep() & remove return.

microbit-carlos:
The `@note` and `@code` is copy-pasted from the DAL docstrings.
This also removes the `@return` info as this is a void method.

Fixes https://github.com/lancaster-university/codal-microbit-v2/issues/150.


JohnVidler:
Happy to merge this - its good to get it down in the docs :)

Issue #208, opened 2022-07-08
Does bluetooth need to know if the board is a V2.0 or V2.2?


microbit-carlos:
Board ID for BLE added in PR https://github.com/lancaster-university/codal-microbit-v2/pull/114

Actually at this point in time CODAL it's not recognising V2.2 on BLE init, only V2.0:
https://github.com/lancaster-university/codal-microbit-v2/blob/b587d9c26f5e93a2054ba8b36a6cdcc27b234419/source/bluetooth/MicroBitBLEManager.cpp#L440-L450

I don't know exactly how this string is used, but it's all processed in the `MicroBitBLEManager::init()` method. Is it necessary for this BLE string to distinguish between V2.0 and V2.2? Should we just hardcode a `V2` string?

Without knowing much about how the BLE stuff is set up, is this information provided as a specific "info" service with a characteristic that provides board info (in which case it might be useful to provide more detailed information)?
Or is it part of a string that is broadcasted in some other way? (in which case it's less important, we might care to connecto a "BBC micro:bit V2" device, doesn't need to say "BBC micro:bit V2.0")

Does this string need to be set up in `MicroBitBLEManager::init()`? Or is there a different function that could provide this values? (for example, if it's a characteristic that is read, the DAPLink I2C version check can be done on demand)

The reason this is worth considering is that it might be beneficial to avoid having this DAPLink I2C transaction on uBit object init:
https://github.com/lancaster-university/codal-microbit-v2/blob/b587d9c26f5e93a2054ba8b36a6cdcc27b234419/model/MicroBit.cpp#L219-L221

We are considering to panic when the DAPLink I2C comms fails in ticket https://github.com/lancaster-university/codal-microbit-v2/issues/137, and for example, if the board is in MAINTENANCE mode and the programme running doesn't use BLE or datalogging, it's better to just run as normal instead of panicking.

microbit-carlos:
Related to:
- https://github.com/lancaster-university/codal-microbit-v2/issues/137

martinwork:
@microbit-carlos The version is used only for a suffix on the model name in the device information service, for the benefit of clients of the BLE services. The question is whether BLE clients need to distinguish V2.0/V2.2, or V2/V3.

The model name currently differentiates V2.0 from newer versions. See the question about future versions in #77. The idea was also discussed in https://github.com/microbit-foundation/codal-microbit/issues/67 (private), and in https://github.com/lancaster-university/microbit-dal/pull/373.

[...]

Issue #137, opened 2021-07-21
Data logging programme hangs if DAPLink does not have micro:bit specific features


microbit-carlos:
This MakeCode programme hangs at the data logging block, so the micro:bit only shows the sad face image, it never reaches the beating heart:

![Screenshot2021_07_20_110233](https://user-images.githubusercontent.com/29712657/126449930-dd7851a3-dcab-4abe-b90e-3e9e0e3ba56a.jpg)

All micro:bits from factory include the I2C functionality, but if somebody builds DAPLink from source they can build the version without it (technically, this is the only version merged in DAPLink upstream right now, the V2 functionality is still in a PR), or users could be using J-Link instead:
https://www.segger.com/products/debug-probes/j-link/models/other-j-links/bbc-microbit-j-link-upgrade/

This DAPLink image can be used for testing:
[kl27z_microbit_if.hex.zip](https://github.com/lancaster-university/codal-microbit-v2/files/6853627/kl27z_microbit_if.hex.zip)

JohnVidler:
Ideally we should panic rather than hang if the i2c comms for logging data are not working.

Some kind of sensible default timeout should be set in CODAL to stop this a forever-wait.

JohnVidler:
May be related to #191 once we implement the panic code.

microbit-carlos:
We need to consider that MAINTENACE mode in DAPLink doesn't have I2C features,.

So, if we panic when we fail to establish I2C comms, then a micro:bit programme would panic in MAINTENACE mode during `uBit.ini()`, which we don't want, as it should only panic at the point the user is using DAPLink I2C specific features like datalogging:
https://github.com/lancaster-university/codal-microbit-v2/blob/b587d9c26f5e93a2054ba8b36a6cdcc27b234419/model/MicroBit.cpp#L219-L221

In this case we would either have:
a) Bubble up error codes and figure out at what layers to panic
    - In this example it cannot panic inside `power.getVersion()`, but that is a user facing uBit API.
b) Remove this board ID check from `uBit.init()`, and panic at the layer that first detects DAPLink is not responding to I2C.
    - https://github.com/lancaster-university/codal-microbit-v2/issues/208
c) Other?

martinwork:
Couldn't the log data call timeout, and not log the data, From memory, it may be that it isn't actually hanging, but the timeout becomes very long (say 20 * 20 * 10s) when I2C is not responding.

JohnVidler:
[...]

PR #255, opened 2023-01-27
Simplify the BLE - Device Info - Model Number to only indicate V2.

microbit-carlos:
Before PR #114 the string was "BBC micro:bit V2.0", with that PR it was updated to indicate "BBC micro:bit V2.0" for board ID 0x9904 and "V2.X" to any unknown board ID (currently V2.2 boards would still appear as "V2.X").

There is no need for BLE apps or devices to distinguish between different V2 versions, just as we currently don't differentiate between V1.3 and V1.5.

Hardcoding this in CODAL to be "BBC micro:bit V2" also removes an I2C call to the interface chip on uBit.init(), which help us better support unofficial interface chip firmwares, like Segger's J-Link.

Fixes https://github.com/lancaster-university/codal-microbit-v2/issues/208

Related to https://github.com/lancaster-university/codal-microbit-v2/issues/137

### DIFF
--- a/inc/bluetooth/MicroBitBLEManager.h
+++ b/inc/bluetooth/MicroBitBLEManager.h
@@ -123,7 +123,7 @@ class MicroBitBLEManager : public CodalComponent
       * bleManager.init(uBit.getName(), uBit.getSerial(), uBit.messageBus, true);
       * @endcode
       */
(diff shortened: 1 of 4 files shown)


JohnVidler (approved):
This all looks good to me, lets merge this in for *.49

PR #261, opened 2023-02-02
CI: Add workflow to calculate build size diff before & after PR/commit.

microbit-carlos:
This PR uses Google's Bloaty McBloatface to analyse the MICROBIT elf files for memory consumption.

It also builds the PR base commit, or the commit parent, to do a diff and see the additional memory consumed or saved, and adds both analyses to the Action work summary.

And in the case of being a PR (like this one) it will post the diff as a comment.
To show this diff, the initial commit in this PR will contain a bit of code in MicroBit.cpp to create a 4000 byte array and a loop to write some data into it. So, the first bot comment will show the PR as increasing the .bss and .text sections.
I will then force push an update removing this code changes and the bot will post another comment showing the PR didn't increase the build size. 

Screenshot of a summary added to the each action run:

image

### DIFF
--- a/.github/workflows/size-diff.yml
+++ b/.github/workflows/size-diff.yml
@@ -1,0 +1,129 @@
+name: Size diff
+
+on:
+  pull_request:
+    branches: '*'
+  # We can consider turning this off pushes if we wanted to reduce CI minutes
+  push:
+    branches: '*'
+
+jobs:
+  size-diff:
+    name: Run Bloaty
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


microbit-carlos:
As this PR was created from a fork, the default GH token created doesn't have rights to add a comment into this PR, however we can see that a failure to push a comment doesn't fail the CI run.

I'll close this PR in favour of https://github.com/lancaster-university/codal-microbit-v2/pull/262, which was created from a branch in this repo, and will be able to show the PR comment functionality.

PR #262, opened 2023-02-02
CI: Add workflow to calculate build size diff before & after PR/commit.

microbit-carlos:
This PR uses Google's Bloaty McBloatface to analyse the MICROBIT elf file for memory consumption.

It also builds the PR base commit, or the commit parent, to do a diff and see the additional memory consumed or saved.

In the case of being a PR from a branch in this repository (not forks) it will post the diff as a comment as well.

To show an example diff in this PR (instead of an output just showing "no change in size"), the initial commit will contain a bit of code in MicroBit.cpp to create a 4000 byte array and a loop to write some data into it. So, the first bot comment will show the PR as increasing the `.bss` and `.text` sections.
I will then force push an update removing this code and the bot will post another comment showing a zero size increase. 

Both bloaty analysis outputs are added to to the Action work summary, so even PRs from forks will have this info (even if they don't get a PR comment):

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,11 +43,10 @@ jobs:
           repository: 'lancaster-university/microbit-v2-samples'
       # We need to use the checkout action (instead of build.py cloning the
       # repository) so that on PRs we can get the commit from the PR merge
-      - name: Clone the this repository in the libraries folder
+      - name: Clone this repository in the libraries folder
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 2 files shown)


microbit-carlos:
Okay, so now I'll force push without the test C++ code to make the build bigger and the bot should comment again indicating the build size for this PR hasn't changed.

microbit-carlos (commented):
I've tweaked a bit the order of the workflow steps and how the codal.json file is modified and this PR is now ready to be merged 🎉

PR #274, opened 2023-02-16
CI: Build MakeCode (with and without docker) on each commit.

microbit-carlos:
Doesn't deploy anything, just checks that it can build the targets correctly (DAL, CODAL, with and without BLE).

One job builds with the Docker image defined in `pxt-microbit/pxtarget.json` and another job with the locally installed toolchain.

This workflow needs a secret to be added to the repo named `YOTTA_GITHUB_AUTHTOKEN` with a **read-only** fine grained token.

### DIFF
--- a/.github/workflows/makecode.yml
+++ b/.github/workflows/makecode.yml
@@ -1,0 +1,94 @@
+name: Build MakeCode
+
+# MakeCode has to git clone/checkout this repo commit SHA, so only run this
+# workflow on pushes, as PRs generate a "merge commit" that is not clonable.
+on:
+  push:
+    branches: '*'
+
+jobs:
+  build-makecode:
+    strategy:
+      matrix:
+        # One job builds with the local toolchain, the other with Docker
+        pxt-flags: ["PXT_NODOCKER=1", ""]
+      fail-fast: false
+    name: Build MakeCode ${{ matrix.pxt-flags && '(nodocker)' }}
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Setup Python 3.7
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.7'
+      - name: Setup arm-none-eabi-gcc v10.3
+        if: ${{ matrix.pxt-flags }}
+        uses: carlosperate/arm-none-eabi-gcc-action@v1
+        with:
+          release: 10.3-2021.10
+      - name: Install Yotta, Ninja v1.10 & CMake v3.22 via PyPI
+        if: ${{ matrix.pxt-flags }}
+        # MarkupSafe < 2.0 needed for Yotta
+        run: python -m pip install MarkupSafe==1.1.1 ninja==1.10.2.2 cmake==3.22.1 yotta
+      - name: Install srecord
+        if: ${{ matrix.pxt-flags }}
+        run: |
+          sudo apt update
+          sudo apt install srecord
+      - name: Setup Node.js v16
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16.x
+      - name: Clone the pxt-microbit repo
+        uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


microbit-carlos:
As the workflow is configured not to build on PRs, only on git pushes, there is no job running in this repo, but the output can be seen in the commit from my fork:

https://github.com/microbit-carlos/codal-microbit-v2/actions/runs/4193614490

microbit-carlos:
@JohnVidler the workflow here is a bit more involved than other, do you want to review it or happy to merge it as is?

.github/workflows/makecode.yml
```diff
@@ -0,0 +1,94 @@
+name: Build MakeCode
+
+# MakeCode has to git clone/checkout this repo commit SHA, so only run this
+# workflow on pushes, as PRs generate a "merge commit" that is not clonable.
+on:
+  push:
+    branches: '*'
```
JohnVidler, line 7: Do we want this to run on _all_ branch pushes, or just `master` ?

.github/workflows/makecode.yml
```diff
@@ -0,0 +1,94 @@
...
+          python-version: '3.7'
+      - name: Setup arm-none-eabi-gcc v10.3
+        if: ${{ matrix.pxt-flags }}
+        uses: carlosperate/arm-none-eabi-gcc-action@v1
+        with:
+          release: 10.3-2021.10
+      - name: Install Yotta, Ninja v1.10 & CMake v3.22 via PyPI
+        if: ${{ matrix.pxt-flags }}
+        # MarkupSafe < 2.0 needed for Yotta
+        run: python -m pip install MarkupSafe==1.1.1 ninja==1.10.2.2 cmake==3.22.1 yotta
```
JohnVidler, line 31: Is MarkupSafe required? I don't see it being used in the actual action?

.github/workflows/makecode.yml
microbit-carlos: All branches, otherwise we'll merge something to master without knowing if it works in MakeCode.

.github/workflows/makecode.yml
microbit-carlos: It's required to be lower than version 2.0 for Yotta, as it's brought as a dependencies without an upper version limit.

.github/workflows/makecode.yml
JohnVidler: I might be wrong, but isn't the `branches` selection here the _destination_ branch, not the source? So we'd just need `master` ?

.github/workflows/makecode.yml
JohnVidler: Ah, ok, so I guess the rest just auto-install with yotta. Gotcha, that's fine then.

.github/workflows/makecode.yml
microbit-carlos: Yes, so if the workflow is configure to only run on "master" and somebody is working in a feature branch this workflow will not run until it is merged into master (also remember that it doesn't run on PRs either). So a feature branch could be breaking MakeCode but we would only find out after it's been merged in the master branch, as that would be the first time this workflow would run.

JohnVidler: Ah, sorry I misread this as applying to PRs, Ok, happy to merge then.

microbit-carlos: Yeah, the mean problem with PRs is that GitHub will create a merge commit to test the PR as it'd be when merged with the target branch. And this commit hash is not really pushed anywhere, so it cannot be "git cloned" or "git checkout" by build.py, MakeCode, nor MicroPython, and it's only accessible when using the `actions/checkout` in a step in the workflow.

So cannot run this workflow on PRs as MakeCode has no way to access that codal commit hash.

Issue #260, opened 2023-01-31
Clarification on Transmit Power Values 


HemanthSabbella:
The setTranmistPower allows you to set the TX power from 0-7, corresponding to -30dBm to +4dBm. The source code surprisingly has ten levels, though the documentation only has 0-7 levels [(link](https://github.com/lancaster-university/codal-microbit-v2/blob/master/source/MicroBitRadio.cpp)).

Further, the nRF52833 states that the lowest programmable limit is -20 dBm ([link](https://www.nordicsemi.com/-/media/Software-and-other-downloads/Product-Briefs/nRF52833-SoC-Product-Brief.pdf)). The microbit: firmware documentation ([link](https://lancaster-university.github.io/microbit-docs/ubit/radio/)) mentioned power = 0 is equivalent to -30dBm. 

The mapping is unclear and needs clarification.

microbit-carlos:
Thanks for raising this @HemanthSabbella.

Having a look, we have for DAL: [microbit-dal/source/bluetooth/MicroBitBLEManager.cpp#L80](https://github.com/lancaster-university/microbit-dal/blob/602153e9199c28c08fd2561ce7b43bf64e9e7394/source/bluetooth/MicroBitBLEManager.cpp#L80)
```cpp
const int8_t MICROBIT_BLE_POWER_LEVEL[] = {-30, -20, -16, -12, -8, -4, 0, 4};
```

With:
[microbit-dal/inc/drivers/MicroBitRadio.h#L67](https://github.com/lancaster-university/microbit-dal/blob/602153e9199c28c08fd2561ce7b43bf64e9e7394/inc/drivers/MicroBitRadio.h#L67)
[microbit-dal/source/drivers/MicroBitRadio.cpp#L296](https://github.com/lancaster-university/microbit-dal/blob/602153e9199c28c08fd2561ce7b43bf64e9e7394/source/drivers/MicroBitRadio.cpp#L296)
```cpp
#define MICROBIT_RADIO_DEFAULT_TX_POWER         6
```
So the there are 8 levels, default power would be 0 dBm, min -30, max 4dBm.

And for CODAL:
https://github.com/lancaster-university/codal-microbit-v2/blob/bf0bb964fbacbc44bf2a1faff09c211a47a1285b/source/MicroBitRadio.cpp#L35
```cpp
const uint8_t MICROBIT_RADIO_POWER_LEVEL[] = {
  0xD8,  // -40 dBm
[...]

PR #275, opened 2023-02-16
radio: Reduce power level options to match V1 DAL.

microbit-carlos:
As discussed in:
- https://github.com/lancaster-university/codal-microbit-v2/issues/260

This change more closely matches the input values for `MicroBitRadio::setTransmitPower()` that micro:bit V1 DAL has, so code trying to achieve the max possible value (`uBit.radio.setTransmitPower(7)`) works correctly in both V1 & V2.

I'll leave the PR in draft state until the discussion in https://github.com/lancaster-university/codal-microbit-v2/issues/260 is completed.


microbit-carlos:
Discussed in the call that we should match the value ranges for V1, so this PR is ready for review @JohnVidler

PR #276, opened 2023-02-16
CI: Build MicroPython on each commit.

microbit-carlos:
As it only builds on pushes the PR doesn't produce a build, but this build from my fork can be seen as an example:
https://github.com/microbit-carlos/codal-microbit-v2/actions/runs/4196282038


microbit-carlos (commented):
This is ready, so merging.

PR #282, opened 2023-02-28
CI: Enable manual triggers for size diff workflow.

microbit-carlos:
With this update to the Bloaty CI workflow we can trigger runs online via the GitHub website.

For example, once this PR is merged, to compare the main branch HEAD with the v0.2.43 commit (5785bb0), we can go to the Bloaty workflow view and trigger a new run with this form:
https://github.com/lancaster-university/codal-microbit-v2/actions/workflows/size-diff.yml
image

And it will calculate size difference between that commit and the main branch HEAD.
To show what that would look like we can have a look at this run from my fork: https://github.com/microbit-carlos/codal-microbit-v2/actions/runs/4295091495
image

### DIFF
--- a/.github/workflows/size-diff.yml
+++ b/.github/workflows/size-diff.yml
@@ -5,6 +5,15 @@ on:
     branches: '*'
   push:
     branches: '*'
+  workflow_dispatch:
+    inputs:
+      old-commit:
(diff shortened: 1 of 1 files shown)


microbit-carlos (commented):
This is ready and only affects the CI, so I'll merge it.

PR #287, opened 2023-03-13
MicroBitUtilityService - replaces PR 178

martinwork:
A fresh branch from current master.
Replaces https://github.com/lancaster-university/codal-microbit-v2/pull/178

TODO: How should the service and characteristic UUIDs determined?
This currently uses the micro:bit base UUID plus 1 and 2.

TODO: MICROBIT_ID_UTILITY - ID definitions have moved into codal-core

MicroBitUtilityService.h/cpp - BLE service providing replies to simple client requests
MicroBitUtilityTypes.h - Details of the service requests that support reading MicroBitLog data
MicroBitCompat.h - Define MICROBIT_ID_UTILITY for service events
MicroBitConfig.h - Macros to control creation in pairing / application mode, default pairing mode only.
MicroBit.cpp - Create service according to config macros. To create in BLEManager::init would need to pass MicroBitLog.
MicroBitBLEChar.cpp - Changed DMESGF to DMESG.


inc/bluetooth/MicroBitUtilityService.h
```diff
@@ -0,0 +1,207 @@
...
+#include "MicroBitEvent.h"
+#include "EventModel.h"
+#include "MicroBitLog.h"
+
+/* Set this to enable an alternative implementation, which creates a fiber to process requests.
+ * The default simply uses an event to trigger processing, and if processRequest() blocks,
+ * a temporary fiber is created as necessary.
+ */
+#ifndef MicroBitUtilityService_FIBER
+#define MicroBitUtilityService_FIBER 0
```
finneyj, line 49: Is there a reason to hold both these implementations in the final revision @martinwork? or shall we just pick one? (e.g. the event based approach?)

martinwork: Yes please pick one - whichever you think is the best idea.

martinwork:
@finneyj I tested it now, using https://github.com/martinwork/microbit-v2-samples/tree/test-mydata.

I had to modify MicroBit.cpp for the panic 071 problem. Would you like me to add a commit for that here? Make a separate PR?

source/bluetooth/MicroBitUtilityService.cpp
```diff
@@ -0,0 +1,385 @@
...
+ * Record the  request, and trigger processing.
+ * Requests are not queued. If a request is actively being processed, ignore the new request, otherwise simply override the previous one.
+ */
+void MicroBitUtilityService::onDataWritten(const microbit_ble_evt_write_t *params)
+{
+    if ( params->handle == valueHandle( mbbs_cIdxCTRL) && params->len > 0 && params->len < sizeof(request_t))
+    {
+        if ( workspace)
+        {
+            if ( workspace->lock)
```
finneyj, line 228: is this safe @martinwork? If I understand correctly, this code will silently ignore incoming requests if a request is already being processed? Is that conventional for BLE apps? Seems quite lossy?

source/bluetooth/MicroBitUtilityService.cpp
```diff
@@ -0,0 +1,385 @@
...
+int MicroBitUtilityService::processLogRead()
+{
+    requestLogRead_t *request = (requestLogRead_t *) &workspace->request;
+    
+    while ( request->batchlen)
+    {
+        if ( workspace->replyState == replyStateClear)
+        {
+            int block = min( request->batchlen, sizeof( reply_t) - offsetof( reply_t, data));
+            int result = log.readData(workspace->reply.data, request->index, block, (DataFormat) request->format, request->length);
```
finneyj, line 359: Does this mean we're very likely to see a sequence of 20 byte read operations from the log @martinwork? That might add quite a bit of latency, compared to one larger read operation?

finneyj:
Generally looks good to me @martinwork! Thanks for the contribution!

I do think we should keep this aligned with the USB based mailbox implementation we drafted for reading the log in due course - just so that we can keep and consistent API for both USB and BLE based transactions. This is already pretty close though, and seeing as we don't want to hold this up any longer, I suggest we tackle that down the line - perhaps by aligning the USB implementation with your BLE one here...

source/bluetooth/MicroBitUtilityService.cpp
martinwork: I believe my thinking was that the each request would expect a reply (both current ones do), and the client wouldn't send multiple requests.

source/bluetooth/MicroBitUtilityService.cpp
martinwork: Won't most of the 20 byte reads be from uBit,log's cache? I suppose there's a fair bit of code for each read to arrive at the memcpy.

martinwork:
Thanks for checking it @finneyj

source/bluetooth/MicroBitUtilityService.cpp
finneyj: I think that's ok - we just need to record somewhere in docs that we don't permit overlapping transactions.

source/bluetooth/MicroBitUtilityService.cpp
finneyj: yes, you're quite right! I forgot we applied the block cache from the file system API into the log reads... All good there then.

I'm not worried about the cache lookup/memcpy cost. It will be insignificant compared to all the BLE/I2C latency...

finneyj:
I'm happy to merge then @martinwork - thanks for the work and quick responses. Unless you have any further thoughts @JohnVidler?

PR #288, opened 2023-03-24
CI: Restrict Python dep versions as workaround to pip install errors.

microbit-carlos:
Currently the MakeCode workflow is failing because a yotta dependency has been updated and subdependencies had clashes with pip version requirements.


microbit-carlos (commented):
Merging to fix the CI failures.

PR #291, opened 2023-03-29
FFT example

microbit-carlos:
Reviving the branch originally developed by Raphael and updated by Joshua.

This is not really mergeable on this state as the CMSIS libs are added as a submodule to this repo and it changes the `float-abi` compiler flag to `hard` (we need to check if we can compile the CMSIS DSP code with `float-abit=softfp`).

Maybe it's worth considering moving this as a "CODAL library" that can also be used as a MakeCode extension, as the `MicroBitAudioProcessor` class is a fairly self-contained `DataSink`.

I'm not sure if I've wired the right DataSource to `MicroBitAudioProcessor` as a I get a lot high frequency values in a relatively quiet room (do we have a DataSource that has filtered out noise?) and the calculated frequencies are twice as expected.
https://github.com/lancaster-university/codal-microbit-v2/blob/e3319f469168b1f3a8ac929de9eba3d413d9501a/samples/fft_example.cpp#L18-L37

Also, it looks like CMSIS needs CMake 3.13 min, and that's a newer version than the Yotta docker image, so the MakeCode builds will fail.


microbit-carlos:
Next steps for this work:
- It's possible I haven't wired things correctly as the calculated frequencies are twice the expected value, so we need to figure out the issue there
- The additional resources used by this feature are too large, 100KBs is a huge chunk of flash, we need to look into how to reduce this
- I believe the FFT calculation time was measure by Joshua at around 10-11 ms, which is not bad, but if we want this to be constantly running in the background we might need it to yield a few times to ensure other things have a chance to run
    - Talk: https://www.youtube.com/watch?v=LzkN4j2DB_w
    - One of the reasons this might be constantly running in the background is the possibility to create an "event block" for MakeCode, so that things can happen when a specific frequency is detected

microbit-carlos:
When using this branch for the first time, the CMSIS_5 git clone might sometimes produce the following error:

```
Downloading CMSIS/Utilities/Win32/PackChk.exe (2.9 MB)
Error downloading object: CMSIS/Utilities/Win32/PackChk.exe (7934974): Smudge error: Error downloading CMSIS/Utilities/Win32/PackChk.exe (7934974e4e929021a7413a609183ce61142c38d9c87e40925a0feec1bc958892): batch response: This repository exceeded its LFS budget. The account responsible for the budget should increase it to restore access.

Errors logged to '/Users/carlos/workspace/tmp/CMSIS_5/.git/lfs/logs/20251218T101326.591298.log'.
Use `git lfs logs last` to view the log.
error: external filter 'git-lfs filter-process' failed
fatal: CMSIS/Utilities/Win32/PackChk.exe: smudge filter lfs failed
```

This can be fixed by using the `GIT_LFS_SKIP_SMUDGE` flag when building the project:

```bash
# First delete the libraries folder so build.py clones all repos from scratch
rm -rf build/
# This flag skips trying to download any of the git LFS files, which we don't need for the FFT
GIT_LFS_SKIP_SMUDGE=1 python build.py
```

Issue #289, opened 2023-03-27
Is it possible to store ~30 bonds rather than just 5?


jaustin:
@martinwork as I understand it, right now we store only 4 bonds and after the 4th bond we overwrite the least-recently-made bond.
( https://github.com/lancaster-university/codal-microbit-v2/blob/6c1433e877e570f9958c1edab77062d4ed9220fd/source/bluetooth/MicroBitBLEManager.cpp#L394 )

Can you easily estimate how much flash we'd need to store ~30 bonds? This would reduce the liklihood of encountering the need to un-bond from the iOS app screen.

martinwork:
Using fd_stat, 
https://github.com/microbit-foundation/codal-microbit-nrf5sdk/blob/ef4662e13875a7b03e7296d7ac24a2b4d231f323/nRF5SDK/components/libraries/fds/fds.h#L226

It looks like a pairing needs 60 bytes, but only consumes 47 or 48 bytes after garbage collection.

So it appears, 15 - 20 pairings is possible within the current FDS space. I don't know if more bonds means more RAM.

I think the need to forget in Settings/Bluetooth also arises if a flash is attempted in pairing mode before pairing.

```
After USB flash
words_used 2
largest_contig 1022
freeable_words 0


After pairing
words_used 59
largest_contig 965
freeable_words 11


After another pairing
words_used 120
largest_contig 904
freeable_words 26


After garbage collection
words_used 94
largest_contig 930
freeable_words 0
```

jaustin:
> So it appears, 15 - 20 pairings is possible within the current FDS space. I don't know if more bonds means more RAM.

@martinwork could you please make a PR for CODAL that allows this larger number of pairings?

>I think the need to forget in Settings/Bluetooth also arises if a flash is attempted in pairing mode before pairing.

I don't understand this - what circumstances would someone be able to try to flash before pairing? Doesn't the app ensure pairing happens first? (And if not, couldn't it do that by trying to read a characteristic that requires bonding to force the phone to bond?)

martinwork:
After running a few of the tests below, I noticed that iOS Settings/Bluetooth was listing both my micro:bits as simply "BBC micro:bit". Forgetting a pairing, would be a chore if there were a class set of pairings.

> I think the need to forget in Settings/Bluetooth also arises if a flash is attempted in pairing mode before pairing.

[...]

PR #299, opened 2023-04-05
MicroBitBLEManager - Move MICROBIT_BLE_MAXIMUM_BONDS to MicroBitConfig

martinwork:
See https://github.com/lancaster-university/codal-microbit-v2/issues/289.

I can't think what I can do to test this.


microbit-carlos:
Yes please, let's make it a built time config option 👍

We should aim to have all flag defaults in `MicroBitConfig.h`, so let's move it there (looks like it's already identified as a BLE config not ported there yet, I think it's fine to start with this one, even if we don't have the others there yet).

For the DAL vs CODAL config in MakeCode, I thought they are separated?

```json
{
    "yotta": {
        "config": {
            "codal": {
                "MICROBIT_BLE_MAXIMUM_BONDS": 20,
            },
            "microbit-dal": {
            }
        }
    }
}
```

martinwork:
@microbit-carlos 
> For the DAL vs CODAL config in MakeCode, I thought they are separated?
Ah I did not know that!

martinwork:
MakeCode project with settings edited to choose 15 bonds and create utility service
https://makecode.microbit.org/_KLihp9h8PXiu

carlosperate (approved):
LGTM 👍 

@finneyj any chance you could have a look at this one in case we can squeze it into today's tag?

finneyj:
Thanks @carlosperate  @martinwork 

Looks good to me! This shouldn't change any default behaviour, but it does feel like we should see what happens if we set a value other than 4 :)  

As I recall, SoftDevice stores its bond data in a dedicate page of FLASH... so my guess is that we're good up until the point that page fills up. 

I'm happy for this to go in as it is low risk, but we should work out what the hard maximum is in due course, and ensure we're not creating undefined behviour by that page overflowing into other areas of FLASH etc.

microbit-carlos:
Yeah, the plan is to be able to expose this a build-time config option, so that we can create a MakeCode project and test if we can store 30 bonds:
- https://github.com/lancaster-university/codal-microbit-v2/issues/289

As we don't have that many devices around, we can get some help testing from teachers and other volunteers, and for that having a MakeCode project to share with them would be quite handy.

finneyj:
cool - thanks carlos. Sounds good!

martinwork:
The bonds are stored by the Nordic SDK in it's FDS file system, for which we have reserved the minimum 2 pages.
https://github.com/martinwork/codal-microbit-nrf5sdk/blob/master/sdk_config/sdk_config.h#L967

I calculated that 20 is the absolute most that might fit without changing the memory map, but 20 might trigger garbage collection, so I suggested 15 as a safer bet. I would guess that exceeding the limit has been tested for 5 devices, but the FDS has never been filled.

PR #302, opened 2023-04-06
Remove old FXOS vestiges.

microbit-carlos:
A couple of FXOS related lines left behind, should not affect the build at all.


microbit-carlos (commented):
This shouldn't have any effect on the compiled output (as shown in the bloaty comment), so merging.

PR #351, opened 2023-07-14
CI MicroPython: Don't checkout submodules recursively.

microbit-carlos:
Is not necessary and as MicroPython contains a lot of submodules with SDKs and similar it takes a long time to pull submodules recursively.

With this change CI build time goes down from 8 to 1 minute.

image


microbit-carlos (commented):
As this is a CI-only change and the respective jobs have finished successfully it doesn't need further review and can be merged directly.

Issue #338, opened 2023-06-15
MakeCode CI failure


microbit-carlos:
Currently the MakeCode CI builds are failing:
https://github.com/lancaster-university/codal-microbit-v2/actions/workflows/makecode.yml
<img width="1254" alt="image" src="https://github.com/lancaster-university/codal-microbit-v2/assets/29712657/ac03852c-f61a-4bbb-93c5-b4191a813ad5">

The error is not very descriptive:

```
building bundled libs/blocksprj
*** yotta target bbc-microbit-classic-gcc@https://github.com/lancaster-university/yotta-target-bbc-microbit-classic-gcc
info: get versions for bbc-microbit-classic-gcc
error: Expecting value: line 1 column 1 (char 0)
INTERNAL ERROR: Error: yotta target bbc-microbit-classic-gcc@https://github.com/lancaster-university/yotta-target-bbc-microbit-classic-gcc: exit code 255
    at ChildProcess.<anonymous> (/home/runner/work/codal-microbit-v2/codal-microbit-v2/node_modules/pxt-core/built/buildengine.js:293:24)
    at ChildProcess.emit (node:events:513:28)
    at maybeClose (node:internal/child_process:1100:16)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:304:5)
```

```
Digest: sha256:48fe950715eabc42c6438de7a5c81f15ccdfb7c79b808b6b2fc2f3c30e733c5c
Status: Downloaded newer image for pext/yotta:update-yotta3
/usr/local/lib/python2.7/dist-packages/jwt/utils.py:8: CryptographyDeprecationWarning: Python 2 is no longer supported by the Python core team. Support for it is now deprecated in cryptography, and will be removed in the next release.
  from cryptography.hazmat.primitives.asymmetric.utils import (
info: get versions for bbc-microbit-classic-gcc
error: No JSON object could be decoded
INTERNAL ERROR: Error: Exit code: 255 from cd built/dockeryt; docker run --rm -v /home/runner/work/codal-microbit-v2/codal-microbit-v2/libs/blocksprj/built/dockeryt/:/src -w /src --env YOTTA_GITHUB_AUTHTOKEN=*** pext/yotta:update-yotta3 yotta target bbc-microbit-classic-gcc@https://github.com/lancaster-university/yotta-target-bbc-microbit-classic-gcc
[...]

PR #358, opened 2023-07-31
CI: Update size diff workflow to let you configure the samples commit.

microbit-carlos:
New additions to the samples repo that use newer APIs will cause build failures in the "size diff" workflow, as it might be configured to compile an older versions of CODAL without those APIs.
This is currently the case, as new asserts have been added to CODAL and the samples repo, so any comparison with older CODAL tags will fail.

So, this PR adds another entry to the form to optionally specify a samples repo commit (bottom one in this screenshot):
image

### DIFF
--- a/.github/workflows/size-diff.yml
+++ b/.github/workflows/size-diff.yml
@@ -14,6 +14,10 @@ on:
         description: '(Optional) New commit to use instead of branch HEAD'
         required: false
         default: ''
+      samples-commit:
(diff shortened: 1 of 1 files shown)


microbit-carlos (commented):
This is only a change to a CI script and has been successfully tested, so go to be merged straight in 👍

The failing tests are due to:
- https://github.com/lancaster-university/codal-microbit-v2/issues/338

PR #359, opened 2023-07-31
CI: Temporarily disable pxt docker build until image becomes available.

microbit-carlos:
As mentioned in:
- https://github.com/lancaster-university/codal-microbit-v2/issues/338#issuecomment-1658767316


microbit-carlos:
As this is a CI-only change that will help us catch issues in the MakeCode builds (since right now it's always red), I'll merge this now, as I'd like to submit a few PRs tomorrow and it'd be good to ensure MakeCode doesn't have any issue with those.

PR #360, opened 2023-08-01
Fix issue when missing motion sensor was crashing instead of panicking.

microbit-carlos:
In summary, when a motion sensor was not working, the `autoDetect()` function was returning a reference to a null pointer.
In this case it looks like the compiler must have been automatically generating the `codal::Accelerometer` and `codal::Compass` copy constructors, and instances of those generic parent classes would end up as `uBit.accelerometer` & `uBit.compass`.

With this PR we are returning an instance of `MicroBitAccelerometer` and `MicroBitCompass`, capable to throw a panic on first usage. Instances of these classes were never included in the builds before, which explains the large size increase for this PR.
The simplest update to these classes increased the build by almost 1KB, but the PR has been slimmed down to try to keep the build small. We'll see in the GitHub Actions PR comment the final size for this version.

I've also found a few places were we can reduce some memory consumption in codal-core, but they need independent review to ensure they don't have any unintended ramifications:
- 152 bytes: https://github.com/lancaster-university/codal-core/pull/162
- 40 bytes: https://github.com/lancaster-university/codal-core/pull/163
- 184 bytes: https://github.com/lancaster-university/codal-core/pull/164
 

Fixes https://github.com/lancaster-university/codal-microbit-v2/issues/213.

### DIFF
--- a/source/MicroBitAccelerometer.cpp
+++ b/source/MicroBitAccelerometer.cpp
@@ -26,7 +26,6 @@ DEALINGS IN THE SOFTWARE.
 #include "MicroBitIO.h"
 #include "MicroBitEvent.h"
 #include "MicroBitCompat.h"
-#include "MicroBitFiber.h"
 #include "MicroBitDevice.h"
 #include "MicroBitI2C.h"
 #include "MicroBitCompass.h"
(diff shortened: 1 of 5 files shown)


source/MicroBitCompass.cpp
```diff
@@ -31,320 +31,55 @@ DEALINGS IN THE SOFTWARE.
...
- * @param id the unique EventModel id of this component. Defaults to: MICROBIT_ID_ACCELEROMETER
- *
- */
-Compass& MicroBitCompass::autoDetect(MicroBitI2C &i2c)
-{
-    /*
-     * In essence, the LSM needs at least 6.4ms from power-up before we can use it.
-     * https://github.com/microbit-foundation/codal-microbit/issues/33
-     */
-    target_wait(10);
```
microbit-carlos, line 63: If the `MicroBitAccelerometer::autoDetect()` already adds a wait, there is no need to wait here as well, which why it was removed.

source/MicroBitAccelerometer.cpp
```diff
@@ -35,321 +34,84 @@ DEALINGS IN THE SOFTWARE.
...
-    /*
-     * In essence, the LSM needs at least 6.4ms from power-up before we can use it.
-     * https://github.com/microbit-foundation/codal-microbit/issues/33
-     */
-    target_wait(10);
-
-    // Add pullup resisitor to IRQ line (it's floating ACTIVE LO)
-    irq1.getDigitalValue();
-    irq1.setPull(PullMode::Up);
-    irq1.setActiveLo();
```
microbit-carlos, line 68: Moved inside the `if (!autoDetectCompleted)` to only wait and configure the GPIO once.

martinwork:
@microbit-carlos Maybe I'm not understanding this, but I didn't think CODAL uBit.accelerometer was ever intended to point to MicroBitAccelerometer. The fall back should be to the base class, codal::Accelerometer, with the panic in Accelerometer::requestUpdate. I thought MicroBitAccelerometer was retained in CODAL for compatabilty with code not using uBit.

microbit-carlos:
> The fall back should be to the base class, codal::Accelerometer, with the panic in Accelerometer::requestUpdate. I thought MicroBitAccelerometer was retained in CODAL for compatabilty with code not using uBit.

Yes, but we'd have to change the generic base class `codal::Accelerometer` to panic in the `requestUpdate` and `configure()` method, which might have an effect in non-micro:bit targets.
I was also under the impression that the CODAL generic error codes and the micro:bit specific 50 & 51 were clashing, which I thought it's why they were added to the MicroBitAccelerometer & MicroBitCompass classes in https://github.com/lancaster-university/codal-microbit-v2/commit/93318d2676a0659c180c05a2fcb11fccc54d6bb6.

microbit-carlos:
@JohnVidler @finneyj with a faulty motion sensor in the current CODAL we end up with a null pointer, so we when trying to execute methods of that instance the micro:bit crashes into an Arm exception handler.

We could fix that by creating instances of one of these two options:
a) Base class `Accelerometer` or `Compass` updated instances that throw a panic on first usage
b) Derived classes `MicroBitAccelerometer` or `MicroBitCompass` that throw a panic on first usage

This PR, for option b), instantiates `MicroBitCompass` & `MicroBitAccelerometer`, which I assumed was the desired effect of https://github.com/lancaster-university/codal-microbit-v2/commit/93318d2676a0659c180c05a2fcb11fccc54d6bb6.

On the other hand we can move the panic to the base class `Accelerometer` & `Compass` (right now they return DEVICE_NOT_SUPPORTED). On a quick test locally, this option would be about 240 bytes smaller (assuming this PR would be merged together with https://github.com/lancaster-university/codal-core/pull/162, which makes the base classes  `configure()` & `requestUpdate()` pure virtual to save 152 bytes). A couple of things about this approach:
- Is this a pattern (throw a panic on base classes) something you'd  be happy to introduce to CODAL?
- Is this going to have an effect to other targets?
- Are we happy to enforce error codes 50 & 51 to be reserved for accelerometer and compass errors in all targets?
    - Right now error code ranges are stipulated by CODAL (like peripheral error), but specific errors (like accelerometer or compass) are defined in each target, so there is a chance these values are already used by a different target to indicate something else?

martinwork:
Ah, I see. I was looking at how it's done in the DAL. If there might be other targets with Accelerometer subclasses that don't override requestUpdate, could a very simple subclass of Accelerometer, which just implements the panic in requestUpdate be a better alternative to MicroBitAccelerometer.

I didn't understand why a panic was added to every MicroBitAccelerometer function 
https://github.com/lancaster-university/codal-microbit-v2/blob/f974d92661e650b8fb998089564e9575fe8190e5/source/MicroBitAccelerometer.cpp
https://github.com/lancaster-university/codal-microbit-v2/blob/master/source/MicroBitAccelerometer.cpp

martinwork:
Yes, sorry @microbit-carlos, I hadn’t understood the result of the changes.

Issue #30, opened 2020-11-17
CODAL Sleep power consumption display oddity


jaustin:
Currently the sleep power is slightly higher than it needs to be because of the way the display sleeps.

Putting in a ticket to track sleep power consumption in CODAL.

I think the most likely is that we're not parking the display pins 'disconnected' appropriately in all cases.

finneyj:
Thanks @jaustin.

I'll take a look next time I find some time.

microbit-carlos:
Possible suspect: https://infocenter.nordicsemi.com/index.jsp?topic=%2Ferrata_nRF52833_Rev1%2FERR%2FnRF52833%2FRev1%2Flatest%2Fanomaly_833_246.html


## [246] System: Intermittent extra current consumption when going to sleep
This anomaly applies to IC Rev. Revision 1, build codes QIAA-Ax0, CJAA-Ax0, QDAA-Ax0.

It was inherited from the previous IC revision Engineering A.
### Symptoms
Extra current consumption in the range of 350 µA when in System On Idle.

### Conditions
A high-speed peripheral (CPU for example) accesses a RAM block which is being accessed by a low-speed peripheral through the DMA bus, with a specific timing, and the high-speed peripheral has higher priority than the low-speed peripheral.

### Consequences
Extra current consumption in System On Idle.

### Workaround
Apply the following code after any reset:
```c
*(volatile uint32_t *)0x4007AC84ul = 0x00000002ul;
```
Workaround consequences: Up to 40 µA current increase when the 16 MHz clock is used.
Parent topic: New and inherited anomalies

microbit-carlos:
## Extra power consumption cause

I've looked into this and found that commenting out this loop reduces the power consumption to around the expected levels:
https://github.com/lancaster-university/codal-microbit-v2/blob/4f2aedcd8e093e7ca9eefd574d4a647884e82505/source/NRF52LedMatrix.cpp#L194-L196


## Why? 🤷‍♂️

The additional power consumption is triggered when the `MODE` (disabled) or the `PSEL` (GPIO number) fields from `GPIOTE.CONFIG[n]` registers are cleared.

<img width="672" alt="image" src="https://github.com/lancaster-university/codal-microbit-v2/assets/29712657/a749aeec-530b-430d-874e-e5ae49c7799a">


I've also noticed that from the 5 channels used (one per column), not all of them have the same impact to the additional power consumption triggered.

And I've also tried these different things without any noticeable improvement:
[...]

PR #366, opened 2023-08-10
Reduce power consumption of LED matrix on power off mode.

microbit-carlos:
As described in:
- https://github.com/lancaster-university/codal-microbit-v2/issues/30#issuecomment-1673717030

### DIFF
--- a/source/NRF52LedMatrix.cpp
+++ b/source/NRF52LedMatrix.cpp
@@ -189,11 +189,14 @@ void NRF52LEDMatrix::disable()
 
     // Disable the timer that drivers the display
     timer.disable();
-    timer.disableIRQ();
 
     // Disable GPIOTE control of the display pins
-    for (int column = 0; column < matrixMap.columns; column++)
-        NRF_GPIOTE->CONFIG[gpiote[column]] = 0;
+    // FIXME: When GPIOTE is disabled here "system off" consumes almost 1mA @ 3V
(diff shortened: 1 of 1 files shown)


source/NRF52LedMatrix.cpp
```diff
@@ -189,11 +189,14 @@ void NRF52LEDMatrix::disable()
 
     // Disable the timer that drivers the display
     timer.disable();
-    timer.disableIRQ();
```
microbit-carlos, line 192: This was removed because it is already done inside `timer.disable()`

microbit-carlos (commented):
It'll be really useful to have this for the factory test programme, so I'll merge this now and we can revisit later.

martinwork:
@microbit-carlos I found this
https://devzone.nordicsemi.com/f/nordic-q-a/94193/nrf52833-about-gpio-default-states-in-system-off-mode

And read in Product Specification... "A GPIO pin input buffer can be disconnected from the pin to enable power savings when the pin is not used as an input"

In disable(), the display pins are set to inputs, but that is with connected input buffer. 

Should they be disconnected (PIN_CNF[PIN] = 2) in setSleep after calling disable, and reconnected at wake up before calling enable?

Looking at setSleep... If the program has disabled the display to use the pins for something else (e.g. ADC), I'm not sure it should be calling disable() on the way to sleep.

Issue #370, opened 2023-08-24
In V2 the BLE "Generic Access" "Device Name" is only `BBC micro:bit`


microbit-carlos:
Should it be `BBC micro:bit V2`?

Basically, in this snippet inside `MicroBitBLEManager::init()`, the `enableBonding` variable is an `init()` parameter only set to `true` when in pairing mode:

https://github.com/lancaster-university/codal-microbit-v2/blob/446765ed7553b0f0d647ba6c9ad61e3b797e1693/source/bluetooth/MicroBitBLEManager.cpp#L265-L272

So basically what we get is

|     |  `Generic Access`->`Device Name` | `Device Information`->`Model Number String`| 
|-----|-----|-----|
| Pairing mode | `BBC micro:bit [abcde]` | `BBC micro:bit V2`|
| Not pairing mode | `BBC micro:bit` | `BBC micro:bit V2`|

This might be correct, but just wanted to double check.

microbit-carlos:
This should be fine.

PR #371, opened 2023-09-05
BLE: Simplify device name string creation.

microbit-carlos:
Saw this when looking at https://github.com/lancaster-university/codal-microbit-v2/issues/370, it's a small change that results in the same string, but it saves a bit of flash with gcc v10.3.

The bloatly CI workflow should post here the size diff.

The `modelVersion` instance is not used anywhere else, `disName` results in the same string and is not used anywhere else either.


JohnVidler:
Looks good to me, lets get this merged in.

PR #407, opened 2024-02-27
CI: Restore building MakeCode with Docker.

microbit-carlos:


microbit-carlos (commented):
As this is simply fixing the CI I'll merge this right away.

Issue #364, opened 2023-08-10
Should we remove `uBit.io.face` alias?


microbit-carlos:
Looks like it was meant to be a workaround to not break early PXT versions:
https://github.com/lancaster-university/codal-microbit-v2/blob/4f2aedcd8e093e7ca9eefd574d4a647884e82505/model/MicroBitIO.h#L416

Only usage I found in a GitHub search is this repo, the samples repo and old forks of MicroPython (it was removed in https://github.com/microbit-foundation/micropython-microbit-v2/commit/bbc70b674d1bfe054ffc1f086ca4d804ae45197d).

GitHub code searches to see its usage:
- [uBit.io.face](https://github.com/search?q=uBit.io.face&type=code)
- [MICROBIT_ID_FACE](https://github.com/search?q=MICROBIT_ID_FACE&type=code)
- [MICROBIT_PIN_FACE_TOUCH](https://github.com/search?q=MICROBIT_PIN_FACE_TOUCH&type=code)
- [ID_PIN_FACE](https://github.com/search?q=ID_PIN_FACE&type=code)

microbit-carlos:
- https://github.com/lancaster-university/microbit-v2-samples/pull/85
- https://github.com/lancaster-university/codal-microbit-v2/pull/410

PR #410, opened 2024-02-28
Remove uBit.io.face as it was a temp additional that is no longer needed.

microbit-carlos:
Fixes https://github.com/lancaster-university/codal-microbit-v2/issues/364


microbit-carlos (commented):
Reviewed on a call, let's remove it 💥

PR #417, opened 2024-03-26
Added definitions to match the Makecode effects parameters

JohnVidler:
As per the title, just a set of definitions for downstream languages to use.


dpgeorge:
Yes, this should work for MicroPython.

microbit-carlos (approved):
Ready to be merged 👍

microbit-carlos:
Added to MakeCode in:
- https://github.com/microsoft/pxt-microbit/pull/5768

dpgeorge:
Actually, this doesn't work for MicroPython because we use C and can't include a C++ header in C code...

PR #418, opened 2024-03-27
Set `uBit.io.logo` to capacitive touch mode by default.

microbit-carlos:
@finneyj this would be the simplest way set the io.logo pin capacitive, do you think it should be behind a build flag as well? If so, what would be the best name for it? `MICROBIT_PIN_LOGO_CAPACITIVE`?
In that case we would still like to make sure the default value is set to capacitive, so the flag would be mostly used to disable this feature?

### DIFF
--- a/model/MicroBit.cpp
+++ b/model/MicroBit.cpp
@@ -116,16 +116,13 @@ MicroBit::MicroBit() :
     }
     */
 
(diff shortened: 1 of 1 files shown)


microbit-carlos (commented):
Reviewed on a call with @finneyj and ready to be merged.

PR #420, opened 2024-04-02
Update Changelog script to include changes in lib dependencies.

microbit-carlos:
It also changes the log format, as `git shortlog` groups the commits by author (or other options, but they always have to grouped) and that can get a bit verbose when also including commits from the other libraries.

Also, an advantage of not grouping by author is that the commits are listed historically.

As part of this PR I've also updated the changelog from v0.2.63, as this is the last tag used for MakeCode live, so it'll be useful to have this info when presenting MS the changelog for deployment.

### DIFF
--- a/.github/tools/gitchanges.py
+++ b/.github/tools/gitchanges.py
@@ -49,9 +49,14 @@ def getTags():
     tags = filter(lambda x: "-" not in x, tags)
     return natsorted(tags,reverse=True)
 
-def getCommitsBetween( tagA, tagB = "HEAD" ):
-    commits = os.popen(f"git -P shortlog {tagA}..{tagB}").read().strip().splitlines()
-    return "\n".join(map( lambda x: x.replace("      ", " - "), commits ))
+def getCommitsBetween( tagA, tagB = "HEAD", dir=None):
+    if dir:
+        original_dir = os.getcwd()
+        os.chdir(dir)
+    commits = os.popen(f'git log --format="%s (by %an)" --no-merges --reverse {tagA}..{tagB}').read().strip().splitlines()
+    if dir:
+        os.chdir(original_dir)
+    return "\n - ".join([""] + commits)
 
 def getRepoURL():
     origin = os.popen("git remote get-url origin").read().strip().split( "github.com/", maxsplit=1 )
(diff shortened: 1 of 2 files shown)


.github/tools/gitchanges.py
```diff
@@ -49,9 +49,14 @@ def getTags():
...
     return natsorted(tags,reverse=True)
 
-def getCommitsBetween( tagA, tagB = "HEAD" ):
-    commits = os.popen(f"git -P shortlog {tagA}..{tagB}").read().strip().splitlines()
-    return "\n".join(map( lambda x: x.replace("      ", " - "), commits ))
+def getCommitsBetween( tagA, tagB = "HEAD", dir=None):
+    if dir:
+        original_dir = os.getcwd()
+        os.chdir(dir)
+    commits = os.popen(f'git log --format="%s (by %an)" --no-merges {tagA}..{tagB}').read().strip().splitlines()
```
microbit-carlos, line 56: There was no way to change `shortlog` to not group commits, so I've replaced it with `git log --no-merges` and in a format that includes the git author name at the end of the commit title.
`%s` -> commit title
`%an` -> commit author name

.github/tools/gitchanges.py
```diff
@@ -49,9 +49,14 @@ def getTags():
     tags = filter(lambda x: "-" not in x, tags)
     return natsorted(tags,reverse=True)
 
-def getCommitsBetween( tagA, tagB = "HEAD" ):
-    commits = os.popen(f"git -P shortlog {tagA}..{tagB}").read().strip().splitlines()
-    return "\n".join(map( lambda x: x.replace("      ", " - "), commits ))
+def getCommitsBetween( tagA, tagB = "HEAD", dir=None):
+    if dir:
+        original_dir = os.getcwd()
+        os.chdir(dir)
```
microbit-carlos, line 55: To be able to easily run this function in other directories I've added the `dir` parameter to temporarily change the current working directory.

.github/tools/gitchanges.py
```diff
@@ -49,9 +49,14 @@ def getTags():
...
+    commits = os.popen(f'git log --format="%s (by %an)" --no-merges --reverse {tagA}..{tagB}').read().strip().splitlines()
```
microbit-carlos, line 56: There was no way to change `shortlog` to not group commits, so I've replaced it with `git log --no-merges`. Reverse was needed because that's the order used by `shortlog`, which I guess makes sense with older commits first, as we read top to bottom.

The format includes the git author name at the end of the commit title, in case we wanted to still provide attribution.
`%s` -> commit title
`%an` -> commit author name

microbit-carlos:
@JohnVidler if possible it'd be great to merge this before releasing the next tag, to be able to include this info in the changelog.

microbit-carlos (commented):
Discussed in a call and good to merge 👍

Issue #240, opened 2022-10-11
`using namespace codal` shouldn't be used in header files


carlosperate:
There are a few header files in the different CODAL repositories that have a `using namespace codal;` line, which forces the namespace into anything that includes it and can cause ambiguity and clashes when integrating with other libraires/SDKs (kind of defeats the point of using a codal namespace to begin with).

I can see maybe a need to have it for `MicroBit.h`, but if we could remove its usage in all the other header files that would be useful.

finneyj:
yes, indeed. I recall this was a necessary evil to get codal-microbit-v2 out before launch. Times were tight!

Correct fix is the explicitly namespace the necessary types in header files to avoid namespace pollution...

finneyj:
p.s. I also agree MicroBit.h is a special case, and leaving one in there does make sense I think, in the interesting of keep user programs simpler and not making breaking changes.

finneyj:
maybe wrapped in a CONFIG option though, so the user can easily opt out should they wish to.

carlosperate:
> maybe wrapped in a CONFIG option though, so the user can easily opt out should they wish to.

Oh yeah, that'd be great!

microbit-carlos:
Fixed with:
- https://github.com/lancaster-university/codal-core/pull/170
- https://github.com/lancaster-university/codal-core/pull/171
- https://github.com/lancaster-university/codal-nrf52/pull/43
- https://github.com/lancaster-university/codal-nrf52/commit/660ee5a8499b69f2c92c2fe401f8bfe3e4fa302f
- https://github.com/lancaster-university/codal-nrf52/pull/54
- https://github.com/lancaster-university/codal-microbit-v2/pull/437

PR #437, opened 2024-07-23
Ensure header files and .cpp files use the codal namespace appropriately

carlosperate:
This is part of:
- https://github.com/lancaster-university/codal-microbit-v2/issues/240

The only thing that hasn't been added to the codal namespace are the Mbed header files, as those are only included by programmes explicitly doing `include <mbed.h>`:
https://github.com/lancaster-university/codal-microbit-v2/blob/96faa8526a4127f6738b2cd5d66595f25ee50b35/inc/compat/mbed.h#L11-L17

This PR uses a new config option added to codal-core in https://github.com/lancaster-university/codal-nrf52/commit/8802eb49140e0389e535cb6160d9080efd951ba7. By default is set to `1`, so that the codal namespace is still added globally in MicroBit.h, but the user can change this config option in case they need to integrate CODAL with other frameworks and avoid clashes in the namespace.


model/MicroBit.h
```diff
@@ -31,7 +31,6 @@ DEALINGS IN THE SOFTWARE.
 #include "CodalDevice.h"
 #include "ErrorNo.h"
 #include "NRFLowLevelTimer.h"
-#include "Matrix4.h"
```
carlosperate, line 34: Removed as it wasn't used.

source/bluetooth/MicroBitUtilityService.cpp
```diff
@@ -35,22 +35,19 @@ DEALINGS IN THE SOFTWARE.
...
 
 #define MICROBIT_ID_UTILITY_PROCESS 0
 
 #define MicroBitUtilityService_SLEEP 10
 #define MicroBitUtilityService_TIMEOUT 10
 
-namespace MicroBitUtility
+typedef enum replyState_t
 {
-    typedef enum replyState_t
```
carlosperate, line 47: This enum is declared and only used inside this cpp file, so no need to declare it inside the `MicroBitUtility` namespace, otherwise with these changes each usage in this file would have to be prepended by the namespace.

Issue #406, opened 2024-02-14
Data log "full" marker is not being written to interface storage


microbit-carlos:
Running this programme on tag v0.2.63, which writes to the data log one byte at a time until the log is full:
```cpp
#include "MicroBit.h"

MicroBit uBit;

int main() {
    uBit.init();

    uBit.display.print("?");

    while (true) {
        uBit.log.logString("X");
        uBit.sleep(1);

        if (uBit.log.isFull()) {
            uBit.display.print("D");
            uBit.sleep(100);
        }
    }
}
```
[MICROBIT.hex.zip](https://github.com/lancaster-university/codal-microbit-v2/files/14284689/MICROBIT.hex.zip)

We should expect the last 4 bytes, which are reserved for this marker, to contain `FULL`:
https://github.com/lancaster-university/codal-microbit-v2/blob/d54ce9b784ef1bba22c075b281bb877abf53ea18/source/MicroBitLog.cpp#L258
https://github.com/lancaster-university/codal-microbit-v2/blob/d54ce9b784ef1bba22c075b281bb877abf53ea18/source/MicroBitLog.cpp#L738-L742

However, the last 4 bytes are left untouched: [MY_DATA.HTM.txt](https://github.com/lancaster-university/codal-microbit-v2/files/14284685/MY_DATA.HTM.txt)

<img width="370" alt="image" src="https://github.com/lancaster-university/codal-microbit-v2/assets/29712657/070c4a64-4a45-4640-a52a-6b43d855c0b8">

jaustin:
This is odd - do you know whether the byte-by-byte case ever wrote 'log full'? Does it still work if logging longer lengths? From MakeCode?

microbit-carlos:
It must have worked at some point, as I believe that's how the "log full" message is meant to be displayed:
- https://github.com/lancaster-university/codal-microbit-v2/issues/140

I haven't tried it in MakeCode yet.

martinwork:
It does get to this point (even with a long string to fill memory more quickly)
https://github.com/lancaster-university/codal-microbit-v2/blob/master/source/MicroBitLog.cpp#L740

But I think the last byte before flash.getFlashEnd() is blocked by the equals in this test
https://github.com/lancaster-university/codal-microbit-v2/blob/master/source/FSCache.cpp#L119
and
https://github.com/lancaster-university/codal-microbit-v2/blob/master/source/FSCache.cpp#L87

[...]

PR #438, opened 2024-07-27
Allow log full FUL to be written to last byte

martinwork:
Fixes https://github.com/lancaster-university/codal-microbit-v2/issues/406

With this program, FUL is in the HTM file, and LOG FULL appears, only after this change. 
```
#include "MicroBit.h"

MicroBit uBit;

int main() {
    uBit.init();

    uBit.display.print("?");

    while (true) {
        uBit.log.logString("1234567890,1234567890,1234567890,1234567890,1234567890\n");
        uBit.sleep(1);

        if (uBit.log.isFull()) {
            uBit.display.print("D");
            uBit.sleep(100);
        }
    }
}
```

image

image

### DIFF
--- a/source/FSCache.cpp
+++ b/source/FSCache.cpp
@@ -84,7 +84,7 @@ int FSCache::read(uint32_t address, const void *data, int len)
 	int bytesCopied = 0;
 
 	// Ensure that the operation is within the limits of the device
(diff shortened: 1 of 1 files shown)


microbit-carlos (approved):
Tested and works great, thanks Martin!

Issue #409, opened 2024-02-27
MakeCode `randint()` not random when Bluetooth enabled


microbit-carlos:
From:
- https://github.com/microsoft/pxt-microbit/issues/5409

Replicable example from:
- https://github.com/microsoft/pxt-microbit/issues/5409#issuecomment-1946206199

```cpp
#include "MicroBit.h"

MicroBit uBit;

int main()
{
    uBit.init();
     microbit_seed_random(); 
     int seed = microbit_random(0x7fffffff); 
     DMESG("random seed: %d", seed); 
     uBit.seedRandom(seed); 
     int randomInt = uBit.random(9);
     uBit.display.print(randomInt);
}
```
```json
{
    "target": {
        "name": "codal-microbit-v2",
        "url": "https://github.com/lancaster-university/codal-microbit-v2",
        "branch": "master",
        "type": "git"
    },
    "config":{
        "MICROBIT_BLE_ENABLED" : 1
    }
}
```

With this example the random numbers are always seeded with the same value. Changing `MICROBIT_BLE_ENABLED` to `1` works.

Possible solution:
- https://github.com/microsoft/pxt-microbit/issues/5409#issuecomment-1946863177

martinwork:
@microbit-carlos 

seedRandom() does not generate a random seed when BLE is running
https://github.com/lancaster-university/codal-microbit-v2/blob/96faa8526a4127f6738b2cd5d66595f25ee50b35/source/MicroBitDevice.cpp#L67

seedRandom is called before BLE is enabled
https://github.com/lancaster-university/codal-microbit-v2/blob/96faa8526a4127f6738b2cd5d66595f25ee50b35/model/MicroBit.cpp#L197

The seed for subsequent calls to seedRandom while BLE is running could be made pseudo random by using:
`    r = random(INT_MAX);
`

The softdevice calls work - they don't seem to need a call to start the RNG peripheral
But I don't see an equivalent of NRF_RNG->TASKS_STOP = 1 to stop the RNG afterwards.
```
        for (int i = 0; i < 4; i++)
        {
            // Wait for a number to be generated.
            uint8_t byte;
            do
            {
              sd_rand_application_bytes_available_get( &byte);
            }
            while( byte == 0);

            sd_rand_application_vector_get( &byte, 1);

            r = (r << 8) | ((int) byte);
        }
```

microbit-carlos:
Thanks Martin, that's really useful!
[...]

PR #448, opened 2024-09-13
Add soft device rng portion to MicroBitDevice::seedRandom().

microbit-carlos:
Otherwise, calling this function when BLE is running results in 0xBBC5EED as always being used as the seed.


microbit-carlos:
@martinwork could you review this PR?

This version is slightly slower, but it might be safer, would you agree?

martinwork:
Elsewhere, SOFTDEVICE_PRESENT and ble_running() have been nested the other way. Would another BLE implementation also block direct access to NRF_RNG, etc?
https://github.com/lancaster-university/codal-microbit-v2/blob/e88ca61bbd1fcad76d20fb9e02f9e5c86e381262/source/MicroBitFlash.cpp#L118


Does using sd_rand_application_bytes_available_get make it safer?

If sd_rand_application_bytes_available_get will eventually say 4 bytes are available, then a simple loop on sd_rand_application_vector_get should eventually succeed.

The only error sd_rand_application_vector_get returns is not enough bytes. In that case it has not written to the buffer.
https://github.com/microbit-foundation/codal-microbit-nrf5sdk/blob/d41d5c7ebe53a1d01935e61d4ffa891e5112e119/nRF5SDK/components/softdevice/s113/headers/nrf_soc.h#L472

Maybe there's a faint to non-existent possibility that an interrupt could steal a byte between sd_rand_application_bytes_available_get and sd_rand_application_vector_get?

I experimented some more, for interest... sd_rand_application_pool_capacity_get returns 64. Immediately after uBit.init() there are 1-3 bytes available. It takes >6500us for bytes_available to be 64. After that, claiming 4 bytes repeatedly takes <20us for the first 17 requests, then each request takes 300-500us.

With a loop like this, I haven't seen count go over 100.
```
      for ( int count = 0; count < 1000; count++)
      {
        if ( sd_rand_application_vector_get((uint8_t*)&r, sizeof(r)) == NRF_SUCCESS)
          break;
      }
```

Assuming any call like sd_rand_application_vector_get will eventually succeed makes me nervous, but I suppose it's just like this.
https://github.com/lancaster-university/codal-microbit-v2/blob/e88ca61bbd1fcad76d20fb9e02f9e5c86e381262/source/MicroBitDevice.cpp#L82

microbit-carlos:
> Elsewhere, SOFTDEVICE_PRESENT and ble_running() have been nested the other way. Would another BLE implementation also block direct access to NRF_RNG, etc?
> 
> https://github.com/lancaster-university/codal-microbit-v2/blob/e88ca61bbd1fcad76d20fb9e02f9e5c86e381262/source/MicroBitFlash.cpp#L118

Yeah, we have discussed in the past exploring other BLE stacks, definitely not in the short term, but it is possible in the future that `ble_running()` could return true and SD not being present.

> Does using sd_rand_application_bytes_available_get make it safer?

Right, not safer than checking the return state of `sd_rand_application_vector_get` in a loop, I was just comparing it as safer than the DAL implementation, which tries to get 4 bytes only once:
https://github.com/lancaster-university/microbit-dal/blob/master/source/core/MicroBitDevice.cpp#L343 

> If sd_rand_application_bytes_available_get will eventually say 4 bytes are available, then a simple loop on sd_rand_application_vector_get should eventually succeed.

Do you mean a loop trying to get one byte at a time instead of trying to get 4 bytes out in one go?

> The only error sd_rand_application_vector_get returns is not enough bytes. In that case it has not written to the buffer. https://github.com/microbit-foundation/codal-microbit-nrf5sdk/blob/d41d5c7ebe53a1d01935e61d4ffa891e5112e119/nRF5SDK/components/softdevice/s113/headers/nrf_soc.h#L472

Ah, or do you mean to try in a loop until success?

```cpp
while (sd_rand_application_vector_get((uint8_t*)&random_value, sizeof(random_value)) != NRF_SUCCESS);
```
> 
> Maybe there's a faint to non-existent possibility that an interrupt could steal a byte between sd_rand_application_bytes_available_get and sd_rand_application_vector_get?

Yes, but I think in those edge cases is probably fine for the default seed to be returned.

martinwork:
>> it is possible in the future that ble_running() could return true and SD not being present. 

I think the original idea was not to burden a no-BLE build with checking ble_running, and that maybe only softdevice precludes using the direct calls. On adding a non-softdevice BLE: if nested the old way it might just work; this way, it would initially revert to doing nothing, and need to be edited even if using NRF_RNG is OK.

>> Ah, or do you mean to try in a loop until success? 

Yes, instead of  a loop trying sd_rand_application_bytes_available_get. It's a bit hard to see in
https://github.com/lancaster-university/codal-microbit-v2/issues/409#issuecomment-2347366765

finneyj:
Where are we up to on this @microbit-carlos  @martinwork? It felt like martin had some unanswered questions?

microbit-carlos:
Ah yes, this was workarounded in MakeCode so the urgency dropped. I'll implement Martin's review comments next week and should be ready for merge after that.

PR #454, opened 2024-11-25
Ensure float operations in `SoundSynthesizerEffects` & `CompassCalibrator`.

microbit-carlos:
Related PR, which includes files in codal-core like `LevelDetectorSPL`, and together with this PR it saves >400 bytes of flash:
- https://github.com/lancaster-university/codal-core/pull/176

_Edit: Also ensured float literals in a couple of other files to avoid implicit double promotion._

### DIFF
--- a/source/MicroBitCompassCalibrator.cpp
+++ b/source/MicroBitCompassCalibrator.cpp
@@ -168,7 +168,7 @@ CompassCalibration MicroBitCompassCalibrator::spherify(Sample3D centre, Sample3D
 
     for (int i = 0; i < samples; i++)
     {
-        int d = sqrt((float)centre.dSquared(data[i]));
+        int d = sqrtf((float)centre.dSquared(data[i]));
 
         if (d > radius)
             radius = d;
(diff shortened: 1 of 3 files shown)


microbit-carlos:
As the PR in codal-core has been merged, I've udpated target-locked.json to point to that commit and that shows the flash saving from having fewer/smaller math functions included in the build.

finneyj:
Looks good - I'm not sure we want to include the target-locked.json though do we @microbit-carlos?

microbit-carlos:
While it is not really necessary for this PR, I actually do prefer when commits going to `codal-microbit-v2` update the `target-locked.json` file **IF** they need a commit from any of its dependencies.

Normally the `target-locked.json` is only updated when a new tag has been created, and that has in some occasions caused `microbit-v2-samples` to have broken builds.
This happens when the main branch is in that in-between state after merging everything into all the codal repositories, but before creating a `codal-microbit-v2` tag with the updated `target-locked.json`. This has also been a bit annoying when trying to go back to old commits for testing or doing a git bisect, and having to guess what the right codal-core/codal-nrf52 commits were that worked with that specific commit from `codal-microbit-v2`.
On the other hand, doing this can create merge conflicts if multiple PRs do that, but I'd probably argue that is easy enough to resolve, and having the main branch always in working state does feel worthwhile.

That doesn't strictly applies to this PR though, as it doesn't really need https://github.com/lancaster-university/codal-core/commit/b395ae364d00597c10b90e0b45aaad0ff15c8495 to compile, so happy to remove it in this case.

microbit-carlos (commented):
Okay, I've reverted the target-locked.json update, so this is ready to be merged 👍

PR #456, opened 2024-11-26
Fix Pi value typo.

microbit-carlos:
I assume this was a simple typo, but since it didn't use the `PI` macro either, I just wanted to double check in case it wasn't @finneyj.


microbit-carlos (commented):
Reviewed with @finneyj in a call and this is good to be merged.

PR #457, opened 2024-12-04
Audio refactor

finneyj:
Not ready to merge.  Raising this PR for review only.


source/MicroBitAudio.cpp
microbit-carlos, line 1: Would `uBit.audio` need a method to change the ADC sampling rate? Or is that located somewhere else?

martinwork:
@finneyj See sample below. Testing by recording silence for 0.1s and looking at the data dumped by serial streamer 

**When an analogue pin is being read, StreamRecording sometimes waits forever** 

Recording with button A hangs after analogue reading is enabled with P0 (cap touch). Recording with button B activates the mic first, and doesn’t hang. A works after a B recording. I think...

When activateMic calls NRF52ADC::activateChannel it doesn’t do anything because the channel has been marked as enabled via dataWanted. NRF52ADC::IsEnabled() is just a flag. It does not mean the channel is actively being checked by an ADC task (one of the enabledChannels count).

deactivateMic doesn’t deactivate the mic ADC channel, but other things might (like starting or stopping the analogue reading). Then the enabled/active difference matters again.


**There seems to be a variable amount of junk at the beginning of a recording**

With trace added, I have seen activateMic seemingly called after the first buffer demux and pull. I cannot repeat that right now - maybe something to do with trace?

Calling setStartDelay in activateMic may be too late, and I think the delay needs to be more than one buffer.

Button B recording seems to get better results by calling activateMic() then setStartDelay(4 or 5) before each recording. 

The start delay required is really a time, so the number of buffers depends on sample rate, and number of other channels.

The first buffer of the first recording after reset is weird because the normalizer is not initialised.


**StreamRecording::play() always sends an empty buffer after the recorded buffers.**


**StreamRecording::play() into SerialStreamer dumps the data but then hangs.**

Everything happens inside the playAsync call to pullRequest, so StreamRecording notifies playLock, before playAsync exits and playLock.wait is called. The sample uses playAsync, which works like play should!


**Sample hex** [MICROBIT.zip](https://github.com/user-attachments/files/18932792/MICROBIT.zip)
A = record, B = activateMic then record, Logo = serial stream, P0 (cap touch) = toggle reading P2 (default is off)

Sample code in microbit-v2-samples "audio-refactor" branch
https://github.com/martinwork/microbit-v2-samples/blob/audio-refactor/source/test-audio-refactor.cpp

I created mw-audio-refactor-merge branches by branching from master, then merging audio-refactor, then branched again to mw-audio-refactor-test to add trace and debug code. e.g. 
https://github.com/martinwork/codal-core/tree/mw-audio-refactor-merge
https://github.com/martinwork/codal-core/tree/mw-audio-refactor-test

martinwork:
I noticed commit [Increase start up delay to 3 frames, to avoid occasional clicks.](https://github.com/lancaster-university/codal-microbit-v2/pull/457/commits/b4514ab83ed21b26a1a1dd545764127335fbbdf5)

And wondered why I saw 1

It was changed in [Reduce sensitivity of microphone to permit detection of 35db](https://github.com/lancaster-university/codal-microbit-v2/pull/457/commits/32a2484e350e16c8cfcfa395a755e763cae35db9)

source/MicroBitAudio.cpp
microbit-carlos: I've created an `audio-refactor` branch in the microbit-v2-samples repository, so that the codal-microbit-v2 CI can pull that and test that it builds correctly (otherwise the main branch fails due to the missing `requestSampleRate()` methods).

I've simply commented out the `requestSampleRate()` calls. @finneyj can you update the branch with the right way to set up the audio recording sampling rate? Not the mixer rate, but microphone sampling.
- https://github.com/lancaster-university/microbit-v2-samples/commit/d360e048da19e4a0ad71f2eb40d9b589aa13e7d8

microbit-carlos:
@finneyj to fix some of the CI issues we need to merge changes in the main branch, do you prefer to rebase or to merge main here?

finneyj:
cool - yes, rebasing here would make sense for CI. If you could do it that would be great - thanks @microbit-carlos!

microbit-carlos:
Okay, so there was one weird merge commit that somehow duplicated the first 8 commits of the branch that was causing weird merge conflicts and bad automatic merges after the rebase.

image

-----

So the first step was trying to clean this branch, which I think I've done correctly in here:

image

https://github.com/lancaster-university/codal-microbit-v2/compare/621ad1f429bee551f0cc049a18fdd0679ad7705e..d9a5a736750651b1e8434fc3cf477a570c9ba02f

microbit-carlos:
Okay, that was much better now, no weird conflicts. The rebase with the main branch should now be complete.

source/MicroBitAudio.cpp
```diff
@@ -55,12 +55,17 @@ MicroBitAudio::MicroBitAudio(NRF52Pin &pin, NRF52Pin &speaker, NRF52ADC &adc, NR
...
         MicroBitAudio::instance = this;
 
+    // Request a periodic callback
+    status |= DEVICE_COMPONENT_STATUS_SYSTEM_TICK;
+
     synth.allowEmptyBuffers(true);
 
     mic = adc.getChannel(microphone, false);
-    adc.setSamplePeriod( 1e6 / CONFIG_MIXER_DEFAULT_CHANNEL_SAMPLERATE );
+    mic->setSampleRate(11000);
```
microbit-carlos, line 64: Should the default rate be a config flag? These flags have been useful when debugging issues in MakeCode as they can be set via pxt.json file.

source/MicroBitAudio.cpp
microbit-carlos: This was done via https://github.com/lancaster-university/codal-microbit-v2/issues/482

source/MicroBitAudio.cpp
microbit-carlos: Thanks Joe!

microbit-carlos:
@finneyj just a reminder that if the target-locked hashes are not updated the CI builds fail as they pull older commit.

microbit-carlos:
Alright! let's merge this for v0.3.0!

PR #465, opened 2025-02-13
NFC to GPIO target.json comprehension issue

Joseph-Tasker:
The microbit library contains a file called target.json which is used to create definitions which interact with the nRF52833 SDK. One of these is "CONFIG_GPIO_AS_PINRESET" which changes the functionality of pin 8 and 9 from NFC antennas to GPIO pins 1 and 2. 
However, the SDK checks if the definition has been defined and not the value it holds. Due to this, NFC is not possible if the value is set to 0 as interpreted, rather the solution is to delete/comment the line so the definition is never defined. This information has been added to the README, but this may not be the intended format of the maintainer.


finneyj:
Thanks @Joseph-Tasker. 

I think you've happened upon a more general issue here, of how we should handle defining CONFIG options for compiler flags that test the presence of a macro rather than its value.

I quite like the textual "save the fellow person" approach, but the text string is quite long, and is specific to this particular issue, and couldn't be reused. 

I wonder if we could have a shorter message, and something more general. e.g. "Enabled. Delete to disable." 

or similar??

Joseph-Tasker:
Thanks for the feedback. I've made a change to reflect the feedback where the string now states:

"ENABLED - Delete entire line to Disable"

I believe the use of the work 'entire' makes it clear that the whole line needs to be removed. After asking a few other people they also agreed as they said they world remove the string rather than the whole line. This should also be useable so that other CONFIG options with the same issue can use this line.

Regarding the issue, would it be preferable if I were to make a new request fixing them or to continue this branch and find any other variables with this issue. The one problem is that I can only find these issues when the libraries have been built which I can do in my other projects.

I hope this is the solution you are looking for.

PR #466, opened 2025-02-19
Small optimisation, save a bit of memory from MicroBitIO.

microbit-carlos:
Display and buttons pull configuration is done in their own components.
Moving from a ManageBuffer to a plain array saves another 100-ish bytes.
In total with arm-none-eabit-gcc v10.3 this PR saves 174 bytes of flash and moves out the heap allocation for the ManageBuffer.


finneyj:
Thanks @microbit-carlos! looks good to me.

Are you happy to merge this?

microbit-carlos:
Great, I'll merge this then 👍

PR #473, opened 2025-03-25
ci: Also build MakeCode & MicroPython on PRs.

microbit-carlos:
The main issue we had before is that on PRs GitHub creates a merge commit of the PR branch merged into main, and that git commit is not really pushed anywhere. So, when MicroPython and MakeCode try to git clone codal-microbit-v2 they get a git sha that does not exit in the upstream repo.

In MicroPython this is solved by using the action/checkout into the MicroPython codal/libraries folder.
In MakeCode it is solved by checking if the workflow is run from a PR, and in that case it just clones the PR branch, not the merge commit. That does mean that the MakeCode workflow does not run the test in the merge commit, but at least we'll have the results from the submitter's branch state.

### DIFF
--- a/.github/workflows/makecode.yml
+++ b/.github/workflows/makecode.yml
@@ -1,9 +1,9 @@
 name: Build MakeCode
 
-# MakeCode has to git clone/checkout this repo commit SHA, so only run this
-# workflow on pushes, as PRs generate a "merge commit" that is not clonable.
 on:
   push:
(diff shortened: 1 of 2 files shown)


microbit-carlos (commented):
Okay, the push and PR workflows pass and I've checked that they are picking up the expected commit hashes in each case.
I'll merge this to be able to use it right away.

Issue #464, opened 2025-02-04
uBit.io.logo.isTouched() doesn't work the first time after reset


martinwork:
Arising from support ticket https://support.microbit.org/helpdesk/tickets/85223 (private)

The code below works first time if I press A (uBit.logo.isPressed) but not B (uBit.io.logo.isTouched).

After the uBit constructors, the uBit.logo TouchButton has configured the uBit.io.logo pin for it's needs but not set its IO_STATUS_TOUCH_IN flag.
https://github.com/lancaster-university/codal-core/blob/master/source/drivers/Button.cpp#L182

IO_STATUS_TOUCH_IN is only set inside isTouched, and isTouched is not called from CODAL.

The first call to uBit.io.logo.isTouched(), sees that IO_STATUS_TOUCH_IN is not set, and creates a replacement for uBit.logo. 

To confirm, insert target_panic(999) before this line.
https://github.com/lancaster-university/codal-nrf52/blob/master/source/NRF52Pin.cpp#L569

```cpp
#include "MicroBit.h"
MicroBit uBit;

void onButtonA(MicroBitEvent e)
{
  if ( uBit.logo.isPressed())
  {
    uBit.display.print('Y');
  }
  else
  {
    uBit.display.print('N');
  }
  uBit.sleep(100);
  uBit.display.clear();
}

void onButtonB(MicroBitEvent e)
{
  if ( uBit.io.logo.isTouched())
  {
    uBit.display.print('Y');
  }
  else
  {
    uBit.display.print('N');
  }
  uBit.sleep(100);
  uBit.display.clear();
}

void onButtonAB(MicroBitEvent e)
{
}

void forever()
{
    while (true)
    {
      uBit.sleep(1000);
    }
}


int main() {
    uBit.init();

    uBit.messageBus.listen( MICROBIT_ID_BUTTON_A,  MICROBIT_BUTTON_EVT_CLICK, onButtonA);
    uBit.messageBus.listen( MICROBIT_ID_BUTTON_B,  MICROBIT_BUTTON_EVT_CLICK, onButtonB);
    uBit.messageBus.listen( MICROBIT_ID_BUTTON_AB,  MICROBIT_BUTTON_EVT_CLICK, onButtonAB);

    create_fiber( forever);

    release_fiber();
    return 0;
}
```

finneyj:
Ah, thanks @martinwork. 

Yes, I think we just call io.logo.isTouched() from uBit.init(), and declare success.

[...]

Issue #113, opened 2021-05-29
Should uBit.logo work when switched to TouchMode::Resistive?


martinwork:
Does uBit.logo work only in capacitative mode or does the code below use the wrong method to switch modes?

After pressing button A to select TouchMode::Resistive, the logo is always pressed.

This example is intended to match the MakeCode code in https://github.com/microsoft/pxt-microbit/issues/3724.

[test-pintouch-v2.zip](https://github.com/lancaster-university/codal-microbit-v2/files/6564661/test-pintouch-v2.zip)

```
#include "MicroBit.h"

MicroBit uBit;

void onButtonA(MicroBitEvent e)
{
    uBit.io.logo.isTouched( codal::TouchMode::Resistive);
}

void onButtonB(MicroBitEvent e)
{
    uBit.io.logo.isTouched( codal::TouchMode::Capacitative);
}

void display()
{
   while (true)
   {
     uBit.display.printChar( uBit.logo.isPressed() ? '1' : '0');
     uBit.sleep(100);
   }
}

int  main() {
    uBit.init();

    uBit.messageBus.listen( MICROBIT_ID_BUTTON_A,  MICROBIT_BUTTON_EVT_CLICK, onButtonA);
    uBit.messageBus.listen( MICROBIT_ID_BUTTON_B,  MICROBIT_BUTTON_EVT_CLICK, onButtonB);

    create_fiber( display);

    release_fiber();
}
```

microbit-carlos:
I can confirm that In this example `uBit.logo.isPressed()` does not work as expected.

`uBit.io.logo.isTouched()` does work, even when called without setting a TouchMode argument (as it is already changed when button A or B are pressed). 


The same can be replicated in MakeCode, as it uses `uBit.logo.isPressed()`: [microbit-logo-touch-change.hex.zip](https://github.com/lancaster-university/codal-microbit-v2/files/8633142/microbit-logo-touch-change.hex.zip)

<img width="774" alt="image" src="https://user-images.githubusercontent.com/29712657/166973317-478b8956-218d-4c10-9c5e-57cb1cb99780.png">

```typescript
input.onButtonPressed(Button.A, function () {
    basic.showString("r")
    pins.touchSetMode(TouchTarget.LOGO, TouchTargetMode.Resistive)
})
input.onButtonPressed(Button.B, function () {
    basic.showString("c")
    pins.touchSetMode(TouchTarget.LOGO, TouchTargetMode.Capacitive)
})
[...]

PR #480, opened 2025-05-08
Fix first use of io.logo.isTouched()

martinwork:
Fixes https://github.com/lancaster-university/codal-microbit-v2/issues/464

Test https://github.com/martinwork/microbit-v2-samples/blob/fix-logo-istouched/source/fix-logo-istouched.cpp

Note: with this change, the first call to uBit.logo.isPressed() doesn't work.
https://github.com/search?q=%2F%5B%5Ecnq%5DuBit%5C.logo%2F&type=code

### DIFF
--- a/model/MicroBit.cpp
+++ b/model/MicroBit.cpp
@@ -122,7 +122,7 @@ MicroBit::MicroBit() :
     #endif
 
     // uBit.logo is a Capacitive TouchButton, ensure the io pin is set capacitive as well
-    io.logo.status |= IO_STATUS_CAPACITATIVE_TOUCH;
+    io.logo.isTouched(TouchMode::Capacitative);
 
     // Add pullup resisitor to IRQ line (it's floating ACTIVE LO)
     io.irq1.getDigitalValue();


finneyj:
Thanks @martinwork - looks good. 

if the first call doesn't work, should we just make an addition call to it in the init() function then?

martinwork:
@finneyj 

`if the first call doesn't work, should we just make an addition call to it in the init() function then?
`

This PR adds a call to uBit.io.logo.isTouched() in init()

I think this fixes uBit.io.logo.isTouched(), but transfers the first-use problem to uBit.logo.isPressed()

Existing code using the old uBit.logo.isPressed() must replace uBit.logo with uBit.io.logo.

finneyj:
Thanks @martinwork. Do we have a sense of which interface MakeCode and micropython are currently using?

martinwork:
I think MakeCode and python changed to io.logo. Is that right @microbit-carlos ?

I did this search - only finds repos I have access to of course!
https://github.com/search?q=%2F%5B%5Ecnq%5DuBit%5C.logo%2F&type=code

microbit-carlos:
Yes, MakeCode is using io.logo, and MicroPython was using it from the beggining (well, originally io.face, but that was updated to io.logo):
- https://github.com/microsoft/pxt-microbit/pull/5749
- https://github.com/lancaster-university/codal-microbit-v2/issues/113#issuecomment-2020933871

finneyj:
great - thanks @martinwork @microbit-carlos.

Looks good to me!

PR #481, opened 2025-05-09
Overwrite default CODAL `uBit.audio.levelSPL` 8-bit conversion range.

microbit-carlos:
codal-core has set the default min and max values to 35 and 100 in: https://github.com/lancaster-university/codal-core/commit/d376f1a851457effaf34058c62bd6c42c5ada6a0

MakeCode uses 52 to 120, and MicroPython 52 to 100:
- https://github.com/microsoft/pxt-microbit/blob/v7.1.31/libs/microphone/microphone.cpp#L8-L9
- https://github.com/microbit-foundation/micropython-microbit-v2/pull/132#issuecomment-1256528663

With the current SoundLevelSPL we don't get values much larger than 100 dB, so it makes sense to make the micro:bit target default to 100 instead of 120.

This was originally intended to go in codal-microbit-v2 v0.2.44, but we never got around to it.


microbit-carlos:
This PR in codal-core fixes double promotion when these values are set up via target.json or codal.json files:
- https://github.com/lancaster-university/codal-core/pull/182

microbit-carlos:
I'll merge this now and make a minor CODAL release for MakeCode.

PR #515, opened 2025-12-17
MicroBitDisplay: move display refresh timer ownership to MicroBit

MannuLambrichts:
# Summary
This pull request updates how the display refresh timer is instantiated and owned within the Codal display stack. Instead of allocating the timer inline inside the `MicroBitDisplay` constructor, the timer is now created in `MicroBit` and passed into `MicroBitDisplay` by reference, aligning it with how other Codal timers (system, adc, touch) are managed.

# Motivation
Previously, `MicroBitDisplay` allocated the display refresh timer using `new` within its constructor. The resulting reference was not retained by `MicroBitDisplay` itself, relying instead on internal storage within `NRF52LedMatrix`. Since that member is private, the timer could not be explicitly managed or cleaned up. Also, `NRF52LedMatrix` does not allocate the timer, so it should not be responsible for maintaining it.
In rare (more advanced) scenarios, this could lead to orphaned timers and memory leaks.

# Implementation details
- The display refresh timer is now instantiated in `MicroBit`, similar to all other timers used by Codal.
- `MicroBitDisplay` receives the timer via reference, making ownership and lifetime explicit.
- The existing `MicroBitDisplay` constructor is preserved for compatibility but is marked as deprecated to discourage further use.
- No functional behaviour of the display refresh logic is changed.

### DIFF
--- a/inc/MicroBitDisplay.h
+++ b/inc/MicroBitDisplay.h
@@ -30,6 +30,7 @@ DEALINGS IN THE SOFTWARE.
 #include "codal-core/inc/types/Event.h"
 #include "NRF52LedMatrix.h"
 #include "AnimatedDisplay.h"
+#include "NRFLowLevelTimer.h"
 
 namespace codal
 {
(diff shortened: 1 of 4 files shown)


finneyj:
Thanks @MannuLambrichts - looks good!

PR #520, opened 2026-06-09
Add accessibility features

finneyj:
Introduce accessibility features as requested in https://github.com/microsoft/pxt-microbit/issues/6848

Should be reviewed and merge in tandem with these PRs:
  - https://github.com/lancaster-university/codal-core/pull/194
  - https://github.com/lancaster-university/codal-nrf52/pull/63

Introduces a fourth PWM module which is used to drive an external WS2812B pixel chain, as commonly used in accessibility hardware for the micro:bit. 

Note: This driver should act transparently to any ongoing bluetooth transactions, as it does not need to disable interrupts.


source/NRF52LedMatrix.cpp
```diff
@@ -400,6 +438,95 @@ int NRF52LEDMatrix::setSleep(bool doSleep)
...
+        {
+            ws = new WS2812B();
+            pwm = new NRF52PWM(NRF_PWM3, *ws, WS2812B_PWM_FREQ);
+
+            pwm->setStreamingMode(true, false);
+            pwm->setDecoderMode(PWM_DECODER_LOAD_Common);
+            pwm->setSampleRate(WS2812B_PWM_FREQ);
+
+            accessibilityPin = (NRF52Pin *) matrixMap.accessibility;
+            accessibilityBuf = ManagedBuffer(25 * 3);
```
microbit-carlos, line 516: All these allocations are happening inside the render method, which is interrupt context. Perhaps they should be moved into the enableAccessibility() method?

source/NRF52LedMatrix.cpp
```diff
@@ -400,6 +438,95 @@ int NRF52LEDMatrix::setSleep(bool doSleep)
...
+        if (pwm == NULL)
+        {
+            ws = new WS2812B();
+            pwm = new NRF52PWM(NRF_PWM3, *ws, WS2812B_PWM_FREQ);
+
+            pwm->setStreamingMode(true, false);
+            pwm->setDecoderMode(PWM_DECODER_LOAD_Common);
+            pwm->setSampleRate(WS2812B_PWM_FREQ);
+
+            accessibilityPin = (NRF52Pin *) matrixMap.accessibility;
```
microbit-carlos, line 515: If `accessibilityPin` is set via `setAccessibilityDisplay()` before it is enabled, this would overwrite that pin configuration previously requested.

source/NRF52LedMatrix.cpp
```diff
@@ -400,6 +438,95 @@ int NRF52LEDMatrix::setSleep(bool doSleep)
     return DEVICE_OK;
 }
 
+/**
+ * Configures an optional WS2812B driven accesibility display for this NRF52LedMatrix.
+ *
+ * @param pin A reference to the pin object to drive the display from
+ * @param numberOfLeds The total number of LEDs to drive. Defaults to 25
+ */
+void NRF52LEDMatrix::setAccessibilityDisplay(NRF52Pin &pin, const uint16_t numberOfLeds)
```
microbit-carlos, line 448: Is the expectation that CODAL consumers have to call this method to change the pin?
If that's the case, will it automatically start outputting the neopixel on `matrixMap.accessibility` (now set to P12) on `uBit.init()` until the MakeCode/MicroPython/user code runs `setAccessibilityDisplay()`?
Should a macro be set as well for compile time configuration?

inc/NRF52LedMatrix.h
```diff
@@ -27,12 +27,14 @@ DEALINGS IN THE SOFTWARE.
...
 #include "CodalConfig.h"
 #include "LEDMatrix.h"
+#include "neopixel.h"
 #include "NRFLowLevelTimer.h"
 
 #define NRF52_LED_MATRIX_CLOCK_FREQUENCY        16000000            // Frequency of underlying hardware clock (must b 1MHz, 2Mhz 4Mhz, 8Mhz or 16MHz)
 #define NRF52_LED_MATRIX_FREQUENCY              60                  // Frequency of the frame update for the display
 #define NRF52_LED_MATRIX_MAXIMUM_COLUMNS        5                   // The maximum number of LEDMatrix columns supported by the hardware.
 #define NRF52_LED_MATRIX_LIGHTSENSE_STROBES     4                   // Multiple of strobe period to use for light sense
+#define NRF52_LED_MATRIX_NEOPIXEL_TIMESLOT      5                   // Timeslot index reserved for accessibilty neopixel use
```
microbit-carlos: Would it be a bit more robust if `NRF52_LED_MATRIX_LIGHTSENSE_STROBES` was `NRF52_LED_MATRIX_MAXIMUM_COLUMNS+1` and `NRF52_LED_MATRIX_NEOPIXEL_TIMESLOT = NRF52_LED_MATRIX_LIGHTSENSE_STROBES+1`?

source/NRF52LedMatrix.cpp
```diff
@@ -49,6 +49,7 @@ static void display_irq(uint16_t mask)
  * The display is initially blank.
  *
  * @param displayTimer The NRF52 timer to use to driver the LEDMatrix display.
+ * @param accessibleDisplay The MicroBitAccessibleDisplay instance to use to drive the accessibility display.
```
microbit-carlos: Is this a leftover from a previous implementation?

source/NRF52LedMatrix.cpp
```diff
@@ -400,6 +438,95 @@ int NRF52LEDMatrix::setSleep(bool doSleep)
...
+
+/**
+ * Define the pixel colour to use for WS2812B attached accessibilty display.
+ * Accessibility needs may require colours other than red for usability.
+ *
+ * @param red The red component of the colour to use.
+ * @param green The red component of the colour to use.
+ * @param blue The red component of the colour to use.
+ */
+void NRF52LEDMatrix::setAccessibilityBaseColour(uint8_t red, uint8_t green, uint8_t blue)
```
microbit-carlos: Similar question to the comment for `NRF52LEDMatrix::setAccessibilityDisplay()`, is this meant to be called by CODAL consumers? If so, there is some time from `uBit.init()` to `setAccessibilityDisplay()` where the display would have the "default colour". For example during pairing mode, which happens during uBit.init() as well.

Should a macro be set as well for compile time configuration?

Secon part of the question, I can't see any default set, so this will be zero initialised? So until the CODAL consumer calls this function we won't see anything on the accessibility display.

microbit-carlos:
@finneyj FYI I've updated the target-lock.json file in this branch to be able to get the CI to build this PR successfully where it can (MicroPython will still fail until we updated it to for v0.3.x).

source/NRF52LedMatrix.cpp
```diff
@@ -104,7 +107,7 @@ void NRF52LEDMatrix::setDisplayMode(DisplayMode mode)
     }
 
     // Determine the number of timeslots we'll need.
-    timeslots = matrixMap.rows;
+    timeslots = matrixMap.rows + 1;
```
microbit-carlos: Does this mean that even if the accessibility display is disabled we'll always have the additional time slot allocated and doing nothing?

inc/NRF52LedMatrix.h
finneyj: Thanks @microbit-carlos. Agree that NRF52_LED_MATRIX_NEOPIXEL_TIMESLOT = NRF52_LED_MATRIX_MAXIMUM_COLUMNS, and that writing as such would indeed be more resilient - thank you.

I think NRF52_LED_MATRIX_LIGHTSENSE_STROBES is independent though, and has no direct computaitonal relationship to the other constants....

source/NRF52LedMatrix.cpp
finneyj: Yep! great spot @microbit-carlos. We should delete this line. It is indeed just a leftover from a previous version...

source/NRF52LedMatrix.cpp
finneyj: Indeed - this is suboptimal. We'll look at a way to remove this time period when the accessibility display is not in use.

source/NRF52LedMatrix.cpp
finneyj: Thanks @microbit-carlos. I think that we don't want to create a compile time flag for this one. Having the capability for end users to explicitly change the pin is fine, but I don' think we want anything transparent to sneak into the build at the point in time. We have P12 reserved for this reason (accessibility), so I think we should use it....

source/NRF52LedMatrix.cpp
finneyj: Yep - having a default colour defined through CONFIG options sounds sensible. It also kind-of feels like we might wan to take some advice from people specialising in visual accessibility too, as to what a good default for this colour might be...

source/NRF52LedMatrix.cpp
finneyj: Indeed - we're going to look at ways to close that race condition.

source/NRF52LedMatrix.cpp
finneyj: no, this would only happen once, so the cost is near zero i think?

microbit-carlos: Yeah, this only runs once so the cost is fine, but my concern is that a lot happens inside the constructors and because it happens in interrupt context it could have interrupted a non-reentrant method.

@microbit-petek mentioned you were also discussing the atomicity of `CodalComponent::addComponent()` as that caused issues somewhere else, and I think the same could happen here if other things are moved around and the interrupt is triggered at a different point.
Even if we fix `CodalComponent::addComponent()` we'd need to be confident there aren't issues in other non-reentrarnt methods.

source/NRF52LedMatrix.cpp
microbit-carlos: Right, but if we are outputting data to pin12 there could be accessories with stuff connected anyway, even if we told them not to use the pin.
So even if their extension calls the method to change the pin, or to disable it completely, they will still get spurious data on pin12 that could do who-knows-what to the connected hardware, no?

inc/NRF52LedMatrix.h
microbit-carlos: Yes, you are right didn't understand `NRF52_LED_MATRIX_LIGHTSENSE_STROBES ` when I was when I wrote that comment 👍

source/NRF52LedMatrix.cpp
microbit-petek: I've been thinking more about this question of which pin should by default be used to output the display data.

Previously I was of the opinion that P12 was the correct choice, for the reason of it being reserved as "the accessibility pin" (even if this is not universally recognised), but if we're piping out the display data by default so that accessibility displays work "out of the box" that does mean that the accessibility pin is then not available for any other form of accessibility feature which an accessory may require.

So now I'm thinking that the right choice would actually be to use one of the display pins to output the data (P6 would be my recommendation), meaning that the accessibility pin is still available for any other accessibility feature, including third-party developed ones, without having to first disable the display output.

What do you think @microbit-carlos @finneyj?

source/NRF52LedMatrix.cpp
```diff
@@ -62,6 +63,8 @@ NRF52LEDMatrix::NRF52LEDMatrix(NRFLowLevelTimer &displayTimer, const MatrixMap &
     lightLevel = 0;
     this->mode = mode;
 
+    this->enableAccessibility();
```
microbit-petek: Is there a reason that the `accessibilityEnabled` member is declared as `false` and then set to true in the constructor? It's a minor thing but would it be more explicit to simply have it declared as `true` from the outset?

inc/NRF52LedMatrix.h
```diff
@@ -65,6 +77,49 @@ namespace codal
...
+
+        /**
+         * Define the pixel colour to use for WS2812B attached accessibilty display.
+         * Accessibility needs may require colours other than red for usability.
+         *
+         * @param red The red component of the colour to use.
+         * @param green The red component of the colour to use.
+         * @param blue The red component of the colour to use.
+         */
+	    void setAccessibilityBaseColour(uint8_t red, uint8_t green, uint8_t blue);
```
microbit-petek: Just a thought: if we were to take a `Colour` as an argument rather than separate red, green and blue, it makes defining some built-in colours trivial - is this something worth considering?

source/NRF52LedMatrix.cpp
```diff
@@ -218,13 +221,22 @@ void NRF52LEDMatrix::render()
...
     {
         // We just completed a normal diplay strobe. 
         // Turn off the LED drive to the row that was completed.
         matrixMap.rowPins[strobeRow]->setDigitalValue(0);
     }
+    else if (strobeRow == NRF52_LED_MATRIX_NEOPIXEL_TIMESLOT)
+    {
+        timer.setCompare(0, timerPeriod);
+
+        // Restore the hardware configuration into LED drive mode.
```
microbit-petek, line 237: I think this might need to be conditional based on whether the display is in a `LIGHT_SENSE` mode or not, otherwise we're switching to LED drive mode only to then have to immediately switch back out for light sensing.

@finneyj am I understanding this correctly?

source/NRF52LedMatrix.cpp
```diff
@@ -287,6 +305,26 @@ void NRF52LEDMatrix::render()
...
+        // Put all pins into high impedance mode.
+        for (int row = 0; row < matrixMap.rows; row++)
+            matrixMap.columnPins[row]->getDigitalValue();
+
+        // Update the accessibility display
+        updateAccessibilityDisplay();
+
+        // We can calculate the period of time it will take the PWM to clock out the data in microseconds using:
+        // (300 + 32 * numberOfPixels). he timer is a 16Mhz timer however so we multiple by 16 to get a tickcount.
+        timer.setCompare(0, 16 * (150 + 10 * accessibilityBuf.length()));
```
microbit-petek, line 327: The equation in the code doesn't match the comment. What are each of the parameters in the equation?
My best guess would be `[timer multiplier] * ([transmission overhead] + [byte transmission time] * [number of bytes])`

source/NRF52LedMatrix.cpp
finneyj: Yes, I agree. P6 was the original pin we used here I think....

inc/NRF52LedMatrix.h
finneyj: yes, that's a good idea.

source/NRF52LedMatrix.cpp
finneyj: Hmm - no particular reason! I think being consistent sounds like a good idea.

source/NRF52LedMatrix.cpp
microbit-petek: I've had a chat with Jonny about a default colour, and his steer on this is that it makes sense to make it red, as it's the "samiest". Also, he pointed out that this works nicely for the macro:bit (or is it mega:bit - the big classroom one?)

source/NRF52LedMatrix.cpp
```diff
@@ -287,6 +302,26 @@ void NRF52LEDMatrix::render()
         // Enable the drive pin, and start the timer.
         matrixMap.rowPins[strobeRow]->setDigitalValue(1);
     }
+    else if(strobeRow == NRF52_LED_MATRIX_NEOPIXEL_TIMESLOT)
+    {
+        for (int col = 0; col < matrixMap.columns; col++)
+        {
+            NRF_GPIOTE->CONFIG[gpiote[col]] = 0;
+            timer.timer->CC[col + 1] = timerPeriod * NRF52_LED_MATRIX_LIGHTSENSE_STROBES;
+            matrixMap.rowPins[col]->setDigitalValue(0);
```
microbit-petek, line 315: Here we're indexing into `rowPins` with an index derived from `columns`, and in the next loop we're indexing `columnPins` with a `row` index. I'm guessing these should be the other way around? As far as I understand though driving rows low and setting columns to high-Z is the desired behaviour @finneyj ?

source/NRF52LedMatrix.cpp
```diff
@@ -287,6 +302,26 @@ void NRF52LEDMatrix::render()
         // Enable the drive pin, and start the timer.
         matrixMap.rowPins[strobeRow]->setDigitalValue(1);
     }
+    else if(strobeRow == NRF52_LED_MATRIX_NEOPIXEL_TIMESLOT)
+    {
+        for (int col = 0; col < matrixMap.columns; col++)
+        {
+            NRF_GPIOTE->CONFIG[gpiote[col]] = 0;
+            timer.timer->CC[col + 1] = timerPeriod * NRF52_LED_MATRIX_LIGHTSENSE_STROBES;
```
microbit-petek, line 314: So my understanding is that this is here to essentially just push the column compares way into the future to make sure there's no chance of a compare event triggering while neopixel data is being transferred.
First of all, is the purpose of this mainly just defensive coding, as the GPIOTE channel is disabled on the previous line?
Secondly, I feel like using `NRF52_LED_MATRIX_LIGHTSENSE_STROBES` here confuses the purpose of this code. Would it be reasonable to move the calculation of how long it will take to clock out the data up above this loop and then change this line to
```
timer.timer->CC[col + 1] += 2 * data_clockout_period
```
?
This pushes any compare event out by at twice the expected time it will take to clock out, which as far as I can tell would have the same overall effect.

source/NRF52LedMatrix.cpp
microbit-petek: I've made some changes to address this in 81ab4155f8097058ffc80fd53cd4ca3ae9033b0b. I've ended up having to create a mini state machine, reason being that when we disable the accessibility display we still need one more neopixel timeslot in order to clock out the data to turn off all the LEDs.
I don't love the result, but it's the best I've come up with so far. Feel free to suggest alternatives if you have them :)

Issue #506, opened 2025-10-18
SoundOutputPin disables IRQs and triggers Bluetooth panic 070


martinwork:
Raised in https://github.com/microsoft/pxt-microbit/issues/6629

When SoundOutputPin::updateOutputBuffer disables IRQs, with Bluetooth advertising (open security or paired), softdevice calls NRF_FAULT_ID_SD_ASSERT (panic 070)

https://github.com/lancaster-university/codal-microbit-v2/blob/a69a50fad4863a2559729694d55f9621d30f5b7b/source/SoundOutputPin.cpp#L193

https://github.com/lancaster-university/codal-microbit-v2/blob/a69a50fad4863a2559729694d55f9621d30f5b7b/source/bluetooth/MicroBitBLEManager.cpp#L1648

Changed (between v0.3.1 and v0.3.2) in https://github.com/lancaster-university/codal-microbit-v2/commit/2affe3a99fdb67798cce958512eb001b3524c7bb to fix https://github.com/microsoft/pxt-microbit/issues/6492

Example hex files built from current master, with and without disable/enable_irq. Press A a few times.
[disableirq.zip](https://github.com/user-attachments/files/22985213/disableirq.zip)

<details>
<summary> code </summary>

```
#include "MicroBit.h"

#ifndef MICROBIT_CODAL
#ifdef CODAL_CONFIG_H
#define MICROBIT_CODAL 1
#else
#define MICROBIT_CODAL 0
#endif
#endif

#if MICROBIT_CODAL
#include "Pin.h"
#define PinCompat codal::Pin
#else
#define PinCompat MicroBitPin
#endif


MicroBit uBit;

//////////////////////////////////////////////////////////

uint8_t pitchVolume = 16;

//////////////////////////////////////////////////////////

PinCompat &music_pin()
{
#if MICROBIT_CODAL
  return (PinCompat &) uBit.audio.virtualOutputPin;
#else
  return (PinCompat &) uBit.io.P0;
#endif            
}


void music_pitch( int f)
{
  uint32_t us = floor( 0.5 + 1000000.0 / f);
  if ( music_pin().getAnalogPeriodUs() != us)
      music_pin().setAnalogPeriodUs( us);

  music_pin().setAnalogValue(pitchVolume);
}


void music_stop()
{
  music_pin().setAnalogValue(0);
  music_pin().setAnalogPeriodUs( 20000);
}


void sleep( uint32_t ms)
{
  uBit.sleep( ms);
[...]

PR #526, opened 2026-08-02
BLE: report pc/info and flush the DMESG buffer in the SD fault handler

attila-lendvai:
this was needed to debug the 070 panic: https://github.com/lancaster-university/codal-microbit-v2/issues/506

maybe this depends on https://github.com/lancaster-university/codal-nrf52/pull/64

LLM slop, but looks plausible. it requires review nevertheless by someone who understands the codebase.

---

The fault handler is entered from the SoftDevice fault context, which runs at a priority above the UARTE IRQ. Mask IRQs so the interrupt-driven serial TX is replaced by the robust polled path, then flush the DMESG buffer so the full panic dump is emitted before the panic handler resets the device.

Also include the offending pc and info in the SD assert/memory-access messages, which are essential for diagnosing SoftDevice assertions such as the micro:bit v2 '070' radio-arming deadline violation.

### DIFF
--- a/source/bluetooth/MicroBitBLEManager.cpp
+++ b/source/bluetooth/MicroBitBLEManager.cpp
@@ -71,6 +71,7 @@ DEALINGS IN THE SOFTWARE.
 #include "MicroBitPartialFlashingService.h"
 
 #include "CodalDmesg.h"
+#include "codal_target_hal.h"
 #include "nrf_log_backend_dmesg.h"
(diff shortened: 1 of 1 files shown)


attila-lendvai:
fixed non-debug build.

attila-lendvai:
i have removed a commit that shouldn't have been part of this PR.

